PR #18, opened 2018-05-23
added rmse evaluator

karanshah03:


core/src/main/java/com/expedia/adaptivealerting/core/util/MathUtil.java
```diff
@@ -16,16 +16,32 @@
...
+    public static double getSquare(double value) {
+        return Math.pow(value, 2);
+    }
+
+    public static double getSquareRoot(double value) {
+        return Math.sqrt(value);
+    }
+
+    public static double roundToThree(double value) {
+        return (Math.round(value * 1000) / 1000);
```
williewheeler: This doesn't quite round to three digits. It's still a double and due to how floating point arithmetic works, there is no guarantee that this rounds to three digits.

core/src/main/java/com/expedia/adaptivealerting/core/evaluator/RmseEvaluator.java
```diff
@@ -0,0 +1,26 @@
...
+public class RmseEvaluator implements Evaluator {
+    @Override
+    public double getScore(double[] actual, double[] prediction) {
+        AssertUtil.isTrue(actual.length == prediction.length, "actual.length != prediction.length");
+        int n = actual.length;
+        double resSumSquares = 0.0;
+        for (int i = 0; i < n; i++) {
+            resSumSquares += MathUtil.getSquare(actual[i] - prediction[i]);
+        }
+        return MathUtil.roundToThree(MathUtil.getSquareRoot((resSumSquares / n)));
```
williewheeler: This doesn't actually round the double to three digits (despite appearances)--see below. Also, is there any point to rounding to three here? I don't see it, and it just increases the cost of the calculation.

core/src/main/java/com/expedia/adaptivealerting/core/evaluator/Evaluator.java
```diff
@@ -8,5 +8,7 @@
  */
 
 public interface Evaluator {
-	double computeScore(double observed, double prediction);
+	void update(double observed, double prediction);
```
williewheeler: Please Javadoc your public methods.

core/src/main/java/com/expedia/adaptivealerting/core/evaluator/RmseEvaluator.java
```diff
@@ -13,16 +11,46 @@
 
     private int n;
     private double resSumSquares;
+    private double rmse;
```
williewheeler: No need for this. See below.

core/src/main/java/com/expedia/adaptivealerting/core/evaluator/RmseEvaluator.java
```diff
@@ -13,16 +11,46 @@
...
 
     @Override
-    public double computeScore(double observed, double predicted) {
-        this.resSumSquares += MathUtil.getSquare(observed - predicted);
+    public void update(double observed, double predicted) {
+        double residual = observed - predicted;
+        this.resSumSquares += residual * residual;
         this.n += 1;
-        return MathUtil.getSquareRoot(resSumSquares / n);
+        setRmse(Math.sqrt(resSumSquares / n));
```
williewheeler: This doesn't address the issue I mentioned earlier. During an update there is no need to calculate the RMSE at all so just leave this out. It greatly increases the cost of an individual update (relative to the simple updates to resSumSquares and n).

core/src/main/java/com/expedia/adaptivealerting/core/evaluator/RmseEvaluator.java
```diff
@@ -13,16 +11,46 @@
...
+        this.rmse = 0;
     }
 
     @Override
-    public double computeScore(double observed, double predicted) {
-        this.resSumSquares += MathUtil.getSquare(observed - predicted);
+    public void update(double observed, double predicted) {
+        double residual = observed - predicted;
+        this.resSumSquares += residual * residual;
         this.n += 1;
```
williewheeler: `this.n++;`

core/src/main/java/com/expedia/adaptivealerting/core/evaluator/RmseEvaluator.java
```diff
@@ -13,16 +11,46 @@
...
-        return MathUtil.getSquareRoot(resSumSquares / n);
+        setRmse(Math.sqrt(resSumSquares / n));
+    }
+
+    @Override
+    public double evaluate() {
+        return rmse;
```
williewheeler: Perform the calculation here, as opposed to doing it in the update method and returning the stored value. That way you do the calculation only when it's actually needed.

core/src/main/java/com/expedia/adaptivealerting/core/evaluator/RmseEvaluator.java
```diff
@@ -13,16 +11,46 @@
...
+    @Override
+    public void reset() {
+        this.n = 0;
+        this.resSumSquares = 0;
+    }
+
+    /**
+     * @return the rmse
+     */
+    public double getRmse() {
```
williewheeler: Remove; see above

core/src/main/java/com/expedia/adaptivealerting/core/evaluator/RmseEvaluator.java
```diff
@@ -13,16 +11,46 @@
...
+        return rmse;
     }
+
+    /**
+     * @param rmse
+     *            the rmse to set
+     */
+    public void setRmse(double rmse) {
```
williewheeler: As above I recommend removing this, but even if we kept RMSE as a state variable, we wouldn't want an external setter. Your `rmse` is an internal state variable so it doesn't make sense to provide an interface for external clients to set it.

williewheeler:
Also it appears that a .DS_Store has crept into your PR. Please remove it.

PR #19, opened 2018-05-25
Add PEWMA detector

exp-dasutherland:


core/src/main/java/com/expedia/adaptivealerting/core/detector/PewmaOutlierDetector.java
```diff
@@ -0,0 +1,186 @@
...
+            double initialAlpha,
+            double beta,
+            int trainingLength,
+            double weakThreshold,
+            double strongThreshold,
+            double initValue
+    ) {
+        isBetweenExclusive(initialAlpha, 0.0, 1.0, "initialAlpha must be in the range (0, 1)");
+
+        this.alpha0 = 1 - initialAlpha;  // To standardise with the EWMA implementation we use the complement value.
```
exp-dasutherland: Not sure if the complement terminology is correct.

core/src/main/java/com/expedia/adaptivealerting/core/detector/PewmaOutlierDetector.java
```diff
@@ -0,0 +1,186 @@
...
+        final double dist = abs(value - mean);
+        final double largeThreshold = strongThreshold * stdDev;
+        final double smallThreshold = weakThreshold * stdDev;
+
+        OutlierLevel level = OutlierLevel.NORMAL;
+        if (dist > largeThreshold) {
+            level = OutlierLevel.STRONG;
+        } else if (dist > smallThreshold) {
+            level = OutlierLevel.WEAK;
+        }
```
exp-dasutherland: Very similar logic in `EwmaOutlierDetector`. Might need to think about combining somehow.

core/src/main/java/com/expedia/adaptivealerting/core/detector/PewmaOutlierDetector.java
```diff
@@ -0,0 +1,186 @@
...
+            return 1 - 1.0/this.trainingCount;
+        }
+        return (1 - this.beta*pt)*this.alpha0;
+    }
+
+    public double getMean() {
+        return mean;
+    }
+
+    public double getStdDev() {
```
exp-dasutherland: Not sure if we want to standardise the getters. Could square it and return variance.

core/src/main/java/com/expedia/adaptivealerting/core/detector/PewmaOutlierDetector.java
```diff
@@ -0,0 +1,186 @@
...
+     */
+    public PewmaOutlierDetector(
+            double initialAlpha,
+            double beta,
+            int trainingLength,
+            double weakThreshold,
+            double strongThreshold,
+            double initValue
+    ) {
+        isBetweenExclusive(initialAlpha, 0.0, 1.0, "initialAlpha must be in the range (0, 1)");
```
exp-dasutherland: My reading of the paper had this as exclusive. Not sure if this should also apply to the alpha in EWMA.

williewheeler: You are right that the paper treats it as exclusive. I also share your intuition that we should use the same bounds for EWMA and PEWMA since the models are very closely related.

In this case I think we can just choose how we want to do it, since alpha doesn't appear as a denominator or anything like that. Other treatments of EWMA use different bounds; see e.g., https://www.itl.nist.gov/div898/handbook/pmc/section3/pmc324.htm, where the interval is half-open.

Personally I think we should use the closed interval [0, 1] in both cases. Obviously 0 and 1 are degenerate cases but I don't see any reason to disallow totally ignoring history or totally ignoring currents. I don't feel strongly.

core/src/main/java/com/expedia/adaptivealerting/core/detector/PewmaOutlierDetector.java
williewheeler: Works for me, especially given the bounded interval.

core/src/main/java/com/expedia/adaptivealerting/core/detector/PewmaOutlierDetector.java
williewheeler: I'm OK with aligning EWMA and PEWMA since they are pretty similar. I don't remember if I was returning stdev or variance in that case. It's probably too early to try to standardize this beyond these two detectors since this state is specific to these specific models. (For example with Aquila we do have a midpoint and dispersion, but they aren't necessarily estimates of mean and stdev.)

Whether we return stdev or variance should probably be driven by which internal representation is more directly useful for the calculations, given the online nature of the algo. In other words we wouldn't want to keep moving back and forth with each update. I think that's stdev here given how you are calculating `zt`.

core/src/test/java/com/expedia/adaptivealerting/core/detector/PewmaOutlierDetectorTest.java
```diff
@@ -0,0 +1,115 @@
...
+
+import static com.expedia.adaptivealerting.core.util.MathUtil.isApproximatelyEqual;
+import static java.lang.Math.sqrt;
+import static junit.framework.TestCase.assertEquals;
+import static junit.framework.TestCase.assertTrue;
+
+public class PewmaOutlierDetectorTest {
+    private static final double WEAK_THRESHOLD = 2.0;
+    private static final double STRONG_THRESHOLD = 3.0;
+    private static final int DEFAULT_TRAINING_LENGTH = 30;
```
williewheeler: Nice. We should probably add this training period to the EWMA and PEWMA evaluators themselves. (Not a blocker for the PR; just an idea for future enhancement.)

core/src/test/java/com/expedia/adaptivealerting/core/detector/PewmaOutlierDetectorTest.java
```diff
@@ -0,0 +1,115 @@
...
+    private static final double DEFAULT_ALPHA = 0.05;
+
+    @Test
+    public void pewmaCloseToEwmaWithZeroBeta() throws IOException {
```
williewheeler: Love it

core/src/main/java/com/expedia/adaptivealerting/core/detector/PewmaOutlierDetector.java
exp-dasutherland: Fine with me. I'll change it.

core/src/main/java/com/expedia/adaptivealerting/core/detector/PewmaOutlierDetector.java
exp-dasutherland: Happy to wait for the patterns to emerge.

core/src/main/java/com/expedia/adaptivealerting/core/detector/PewmaOutlierDetector.java
```diff
@@ -0,0 +1,186 @@
...
+
+        return level;
+    }
+
+    private void updateEstimates(double value) {
+        double zt = 0;
+        if (this.stdDev != 0) {
+            zt = (value - this.mean)/this.stdDev;
+        }
+        double pt = (1/sqrt(2*Math.PI)) * exp(-(pow(zt, 2)/2));
```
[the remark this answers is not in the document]
exp-dasutherland: Yeah I like the change. Initially was trying to get it to exactly match the paper but prefer the way this reads.

core/src/main/java/com/expedia/adaptivealerting/core/detector/PewmaOutlierDetector.java
exp-dasutherland: Might wait for a use-case that requires it, in PEWMA we keep stdDev in EWMA we keep variance.

PR #25, opened 2018-05-29
add detector by id lookup/creation

exp-dasutherland:


kafka/src/main/java/com/expedia/adaptivealerting/kafka/util/DetectorUtil.java
```diff
@@ -73,6 +74,30 @@ public static void startStreams(OutlierDetector detector, String appId, String t
...
+            Map<String, OutlierDetector> detectors,
+            Function<String, OutlierDetector> detectorFactory
+    ) {
+        // FIXME Hack, see above
+        if (metricPoint == null) {
+            metricPoint = NULL_METRIC_POINT;
+        }
+        final String metricId = extractMetricId(metricPoint);
+        if (!detectors.containsKey(metricId)) {
+            detectors.put(metricId, detectorFactory.apply(metricId));
```
exp-dasutherland: Not sure if we need to worry about thread safety here.

williewheeler: From my conversation with Magesh it sounded like we are OK on threadsafety.

PR #33, opened 2018-06-01
Added a CSV metric source, and created a pipeline mini-framework.

williewheeler:
For model development and testing, it's useful to be able to connect
outlier detectors to metric sources, and to do this without having to
fire up an underlying messaging infrastructure. I created a basic source/
filter/sink framework in the tools module, which allows us to do exactly
this. I used wrappers for outlier detectors to avoid polluting their
API with what's essentially Java event semantics.

There are a couple of sample pipelines in the
com.expedia.adaptivealerting.samples package.


core/src/main/java/com/expedia/adaptivealerting/core/detector/PewmaOutlierDetector.java
```diff
@@ -92,10 +93,17 @@
      * Local standard deviation estimate.
      */
     private double stdDev;
-
-
+    
+    /**
+     * Creates a new PEWMA outlier detector with initialAlpha = 0, beta = 1.0, weakThreshold = 2.0,
```
djsutho: initialAlpha = 0 -> initialAlpha = 0.5

djsutho: I ave feeling 0. is very high but will have to test and see

PR #50, opened 2018-06-11
This includes following changes

bibinss:
1. Added a new module 'end2end.test' to hold all the e2e integration tests
2. StreamRunner creation logic from all the existing kstream apps have been moved to separate individual classes inorder to easily integrate with e2e tests.
3. ConstantThresholdBasedE2ETestSpec.scala tests the '(anomalous inputs)->(metric-router)->(constant-threshold-detector)->(output-topic)' E2E pipeline

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -55,6 +55,14 @@
         <!-- Maven plugins -->
         <maven.compiler.plugin.version>3.7.0</maven.compiler.plugin.version>
         <maven.jar.plugin.version>3.1.0</maven.jar.plugin.version>
+        <scalatest.version>3.0.3</scalatest.version>
+        <scala.major.version>2</scala.major.version>
+        <scala.minor.version>12</scala.minor.version>
+        <scala.major.minor.version>${scala.major.version}.${scala.minor.version}</scala.major.minor.version>
+        <scala.maintenance.version>2</scala.maintenance.version>
+        <scala-library.version>${scala.major.minor.version}.${scala.maintenance.version}</scala-library.version>
+        <pegdown.version>1.4.2</pegdown.version>
+        <kafka.version>1.1.0</kafka.version>
     </properties>
 
     <dependencyManagement>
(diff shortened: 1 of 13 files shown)


end2end.test/src/test/scala/com/expedia/adaptivealerting/pipeline/test/ConstantThresholdBasedE2ETestSpec.scala
```diff
@@ -0,0 +1,118 @@
...
+      constantThresholdSR.start()
+
+      Then("'metric router' should route the metrics to 'constant threshold outlier detector' model's input topic")
+
+      val intermediateRecords: List[KeyValue[String, MetricPoint]] =
+        IntegrationTestUtils.waitUntilMinKeyValueRecordsReceived[String, MetricPoint](
+          configToProps(constantThresholdTopicConsumerConfig.getConfig(STREAM)),
+          INTERMEDIATE_TOPIC, 2, 15000).asScala.toList // get metricPoints from Kafka's output topic
+      intermediateRecords.size shouldEqual 2
+      intermediateRecords.foreach(i => {
```
ayansen: you could scala pattern matching here

pom.xml
```diff
@@ -37,6 +37,7 @@
         <module>kafka</module>
         <module>samples</module>
         <module>tools</module>
+        <module>end2end.test</module>
```
ayansen: do you need a separate module ? you could also consider adding it as part of the kafka module since you're e2e tests depend on kafka being the event stream

end2end.test/src/test/scala/com/expedia/adaptivealerting/pipeline/test/ConstantThresholdBasedE2ETestSpec.scala
ayansen: something like  intermediateRecords.foreach (record => record.value.metric match { case "latency" => succeed, case "duration" => succeed,  case _ => fail)

end2end.test/src/test/scala/com/expedia/adaptivealerting/pipeline/test/ConstantThresholdBasedE2ETestSpec.scala
```diff
@@ -0,0 +1,118 @@
...
+        List("latency", "duration") contains(i.value.metric) shouldBe true
+      })
+
+      Then("'constant threshold outlier detector' should read records from its topic and " +
+        "write those anomalous records to output topic")
+      val consumerPropAnomalyTopic = configToProps(anomalyTopicConsumerConfig.getConfig(STREAM))
+      consumerPropAnomalyTopic.put("JsonPOJOClass", classOf[AnomalyResult])
```
ayansen: can this be replaced by a serde ?

end2end.test/src/test/scala/com/expedia/adaptivealerting/pipeline/test/ConstantThresholdBasedE2ETestSpec.scala
```diff
@@ -0,0 +1,118 @@
...
+      val outputRecords: List[KeyValue[String, AnomalyResult]] =
+        IntegrationTestUtils.waitUntilMinKeyValueRecordsReceived[String, AnomalyResult](
+          consumerPropAnomalyTopic,
+          OUTPUT_TOPIC, 2, 35000).asScala.toList // get metricPoints from Kafka's output topic
+      outputRecords.size shouldEqual 2
+
+      Then("no other intermediate partitions are created after as a result of topology")
+      val adminClient: AdminClient = AdminClient.create(metricRouterStreamConfig)
+      val topicNames: Iterable[String] = adminClient.listTopics.listings().get().asScala
```
ayansen: get can throw a none.get exception, would recommend adding an assertion that the listings exist before you do a get. this would help your test fail gracefully in case the topics aren't created

end2end.test/src/test/scala/com/expedia/adaptivealerting/pipeline/test/ConstantThresholdBasedE2ETestSpec.scala
```diff
@@ -0,0 +1,118 @@
...
+
+  private def metricRouterStreamRunner = {
+    new MetricRouterStreamBuilder().build(appConfig("metricrouter"))
+  }
+
+  private def constantThresholdDetectorStreamRunner = {
+    new ConstantThresholdOutlierDetectorStreamRunnerBuilder().build(appConfig("constant-detector"))
+  }
+
+  def generateAnomalousMetrics() : List[MetricPoint] = {
```
ayansen: private def ?

end2end.test/pom.xml
```diff
@@ -0,0 +1,201 @@
...
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>adaptive-alerting-kafka</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+
+        <!-- Third-party -->
+        <dependency>
```
ayansen: do you need junit ?

pom.xml
williewheeler: +1

williewheeler: If we ever do a Kinesis wrapper (which I expect we will), I'd prefer to see the Kinesis e2e inside of a kinesis module instead of having a single e2e module that knows about all the different wrappers.

end2end.test/src/test/scala/com/expedia/adaptivealerting/pipeline/test/ConstantThresholdBasedE2ETestSpec.scala
bibinss: serde used here is a generic one - "JsonPojoSerializer" and "JsonPojoDeserializer". We need to pass the class instance type as a parameter for the deserializer to work.

PR #58, opened 2018-06-18
revise anomaly detection sections

Derikulous:
This is to change outlier detection to anomaly detection, and what we used to call anomaly detection as anomaly validation.


williewheeler:
Closed as I simply deleted the offending files

PR #90, opened 2018-07-04
Enhancement/kafka adm main

williewheeler:
The KADM needs to be able to create detectors on demand. To do this we use
the factory pattern. When the KADM starts up, it needs to know the mapping
from detector names to factory classnames. Hence the number of factory-
related classes in this commit.

In the future we want to get rid of the detector-specific KStreams apps
and just use the KADM to host detectors. But for now the system works with
detector-specific apps and I didn't want to break that. So that's all left
alone. As part of the KADM implementation, though, I did externalize the
names of the inbound and outbound topics to config files, and I did this
using the configuration property names inbound-topic and outbound-topic.
To avoid having to handle both "topic" (the old name for the inbound
topic) and "inbound-topic", I just migrated everything over to using the
"inbound-topic" name, and switch the property names over to constants to
establish better control. I've updated the Haystack Terraform to define
both "topic" and "inbound-topic" values for now. Once this AA update is
deployed, we can remove the "topic" definitions from the Haystack
Terraform.

This change also fixes the bad default settings for the EWMA and PEWMA
detectors. EWMA had reversed weak/strong thresholds, and both had settings
that were too aggressive. (Instead of weak/strong of 2/3, I used 3/4. Two
sigmas is way too aggressive for alerting.)

### DIFF
--- a/kafka/src/main/resources/config/base.conf
+++ b/kafka/src/main/resources/config/base.conf
@@ -12,28 +12,26 @@ metric-router {
   streams {
     application.id = "metric-router"
   }
-  topic = "metrics"
+  inbound-topic = "metrics"
+
+  # TODO This doesn't do anything yet, but it will once MetricRouter maps the mpoints. [WLW]
+  outbound-topic = "mapped-metrics"
 }
 
-constant-detector {
+anomaly-detector-manager {
   streams {
-    application.id = "constant-detector"
+    application.id = "anomaly-detector-manager"
+    default.value.serde = "com.expedia.adaptivealerting.kafka.serde.JsonPojoSerde"
   }
-  topic = "constant-metrics"
-}
-
-ewma-detector {
-  streams {
-    application.id = "ewma-detector"
+  factories {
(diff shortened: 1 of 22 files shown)


kafka/src/main/resources/config/base.conf
```diff
@@ -12,36 +12,64 @@ metric-router {
...
+  inbound-topic = "metrics"
+
+  # TODO This doesn't do anything yet, but it will once MetricRouter maps the mpoints. [WLW]
+  outbound-topic = "mapped-metrics"
+}
+
+anomaly-detector-manager {
+  streams {
+    application.id = "anomaly-detector-manager"
+    default.value.serde = "com.expedia.adaptivealerting.kafka.serde.JsonPojoSerde"
```
djsutho: Seems like a serde is provided on the input and output (`jsonMpointSerde` and `jsonAnomalySerde`) so this probably isn't needed otherwise might need to specifiy `JsonPojoClass` as well?

PR #91, opened 2018-07-05
Created a new AnomalyDetectorMapper, intended to replace MetricRouter.

williewheeler:
MetricRouter is deprecated primarily because the "routing" function is no
longer routing as much as it is message splitting and enrichment. But
there were some other things we needed to take care of anyway. The new
AnomalyDetectorMapper takes care of these:

- Separates the core mapping logic (i.e. the splitting/enrichment) from
  the Kafka wrapper
- Consumes Mpoints instead of Haystack MetricPoints
- Produces MappedMpoints instead of Haystack MetricPoints

### DIFF
--- a/kafka/src/main/java/com/expedia/adaptivealerting/kafka/KafkaConfigProps.java
+++ b/kafka/src/main/java/com/expedia/adaptivealerting/kafka/KafkaConfigProps.java
@@ -29,8 +29,14 @@ public interface KafkaConfigProps {
     // ================================================================================
     
     /**
+     * Anomaly Detector Mapper KStreams app name.
+     */
+    public static final String ANOMALY_DETECTOR_MAPPER = "anomaly-detector-mapper";
+    
+    /**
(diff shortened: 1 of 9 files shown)


kafka/src/main/java/com/expedia/adaptivealerting/kafka/KafkaConfigProps.java
```diff
@@ -28,9 +28,15 @@
...
     
+    /**
+     * Anomaly Detector Mapper KStreams app name.
+     */
+    public static final String ANOMALY_DETECTOR_MAPPER = "anomaly-detector-mapper";
+    
     /**
      * Metric Router KStreams app name.
      */
+    @Deprecated
```
djsutho: Nice to add some docs for a deprecated vale.

williewheeler: @djsutho Done, thanks sir.

PR #93, opened 2018-07-05
Enhancement/performance monitor

karanshah03:


anomdetect/src/main/java/com/expedia/adaptivealerting/anomdetect/MonitorDetector.java
```diff
@@ -0,0 +1,59 @@
...
+
+    private Logger LOGGER = LoggerFactory.getLogger(MonitorDetector.class);
+
+    public MonitorDetector(AnomalyDetector detector, PerformanceMonitor perfMonitor) {
+        this.detector = detector;
+        this.perfMonitor = perfMonitor;
+    }
+
+    // FIXME Do we need to return a mapped point here? [KS]
+    public void classify(MappedMpoint mappedMpoint) {
```
williewheeler: Include an `@Override`. That will tell you the answer.

anomdetect/src/main/java/com/expedia/adaptivealerting/anomdetect/MonitorDetector.java
```diff
@@ -0,0 +1,59 @@
...
+        // FIXME Check if we need to classify the mapped point again here. [KS]
+        MappedMpoint mappedPoint = detector.classify(mappedMpoint);
+        AnomalyResult result = mappedPoint.getAnomalyResult();
+        double perfScore = perfMonitor.evaluatePerformance(result);
+        if (perfScore != 0) {
```
williewheeler: Why are we checking the perf score here?

anomdetect/src/main/java/com/expedia/adaptivealerting/anomdetect/PerformanceMonitor.java
```diff
@@ -0,0 +1,75 @@
...
+ * limitations under the License.
+ */
+package com.expedia.adaptivealerting.anomdetect;
+
+import com.expedia.adaptivealerting.core.anomaly.AnomalyResult;
+import com.expedia.adaptivealerting.core.evaluator.RmseEvaluator;
+
+/**
+ * <p>
+ * Performance monitor which generates a result every 100 ticks and resets evaluator. Currently it uses only RMSE
```
williewheeler: The tick count and the evaluator should be configurable. So the documentation should not reference specific values.

anomdetect/src/main/java/com/expedia/adaptivealerting/anomdetect/PerformanceMonitor.java
```diff
@@ -0,0 +1,75 @@
...
+
+    /**
+     * Local tick counter.
+     */
+    private int tickCounter;
+
+    /**
+     * Local RMSE evaluator.
+     */
+    private RmseEvaluator evaluator;
```
williewheeler: This should accept an `Evaluator`, not an `RmseEvaluator`. For now when we create a perfmon we can just pass in an `RmseEvaluator`, but the perfmon itself shouldn't require that.

anomdetect/src/main/java/com/expedia/adaptivealerting/anomdetect/PerformanceMonitor.java
```diff
@@ -0,0 +1,75 @@
...
+
+    /**
+     * Local threshold lookup where we maintain thresholds for different evaluators.
+     */
+
+    public static final int MAX_TICKS = 100;
+    public static final double RMSE_THRESHOLD = 3.0;
```
williewheeler: I assume this RMSE threshold is left over from your earlier impl? Remove that. (Anyway 3 is a very low RMSE for the kinds of series we're looking at--for bookings that would certainly be an amazing RMSE.)

anomdetect/src/main/java/com/expedia/adaptivealerting/anomdetect/PerformanceMonitor.java
```diff
@@ -0,0 +1,75 @@
...
+        double predicted = result.getPredicted();
+        double evaluatorScore = 0;
+        evaluator.update(observed, predicted);
+
+        if (tickCounter >= MAX_TICKS) {
+            evaluatorScore = evaluator.evaluate().getEvaluatorScore();
+            evaluator.reset();
+            resetCounter();
+        }
+        this.tickCounter++;
```
williewheeler: Small nitpick, but we shouldn't both reset the counter _and_ increment it. If the tick counter hits the max, then setting it to 0 _is_ the increment.

anomdetect/src/test/java/com/expedia/adaptivealerting/anomdetect/PerformanceMonitorTest.java
```diff
@@ -0,0 +1,79 @@
...
+        while (testRows.hasNext()) {
+            final PerfMonitorTestRow testRow = testRows.next();
+            final double observed = testRow.getObserved();
+            final double predicted = testRow.getPredicted();
+
+            AnomalyResult result = new AnomalyResult();
+            result.setObserved(observed);
+            result.setPredicted(predicted);
+            double performanceScore = perfMonitor.evaluatePerformance(result);
+            if (count < 100) {
```
williewheeler: Don't hardcode to 100. Use the configured max.

anomdetect/src/test/java/com/expedia/adaptivealerting/anomdetect/PerformanceMonitorTest.java
```diff
@@ -0,0 +1,79 @@
...
+                assertEquals(testRow.getScore(), performanceScore, 0);
```
williewheeler: When comparing floating point numbers you generally don't want a 0 tolerance. This means that the float point representations have to match exactly. In this case you got lucky but in general you want to have a nonzero tolerance. Even 0.001 is fine. Use a constant for it please so you can adjust centrally.

anomdetect/src/test/java/com/expedia/adaptivealerting/anomdetect/PerformanceMonitorTest.java
```diff
@@ -0,0 +1,79 @@
...
+            } else {
+                assertEquals(testRow.getScore(), performanceScore, 0.2);
```
williewheeler: The evaluator shouldn't return 0 all the time. It should return the running RMSE. It's just that it flushes the RMSE every nth tick.

anomdetect/src/main/java/com/expedia/adaptivealerting/anomdetect/PerformanceMonitor.java
```diff
@@ -0,0 +1,74 @@
...
+     */
+    public PerformanceMonitor() {
+        this.evaluator = new RmseEvaluator();
+        resetCounter();
+    }
+
+    public double evaluatePerformance(AnomalyResult result) {
+        double observed = result.getObserved();
+        double predicted = result.getPredicted();
+        double evaluatorScore = 0;
```
williewheeler: Use 0.0 for double, not 0. The JVM casts but just clearer to make it obvious that it's a double.

anomdetect/src/main/java/com/expedia/adaptivealerting/anomdetect/AnomalyDetectorManager.java
```diff
@@ -67,6 +67,10 @@ public AnomalyDetector detectorFor(MappedMpoint mappedMpoint) {
             detector = factory.create(detectorUuid);
             detectors.put(detectorUuid, detector);
         }
+        PerformanceMonitor perfMonitor = new PerformanceMonitor();
+        MonitorDetector monitorDetector = new MonitorDetector(detector, perfMonitor);
+        // FIXME Should this be part of below classify method? [KS]
+        monitorDetector.classify(mappedMpoint);
```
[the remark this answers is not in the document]
williewheeler: In fact, please change this detectorFor() method to accept a Metric instead of an Mpoint just to make this clearer. You'll have to pass in the mpoint.getMetric().

anomdetect/src/main/java/com/expedia/adaptivealerting/anomdetect/AnomalyDetectorManager.java
```diff
@@ -111,6 +107,8 @@ public AnomalyDetector detectorFor(MappedMpoint mappedMpoint) {
     public MappedMpoint classify(MappedMpoint mappedMpoint) {
         notNull(mappedMpoint, "mappedMpoint can't be null");
         final AnomalyDetector detector = detectorFor(mappedMpoint);
+        MonitoredDetector monitorDetector = new MonitoredDetector(detector);
+        monitorDetector.classify(mappedMpoint);
         return detector == null ? null : detector.classify(mappedMpoint);
```
williewheeler: Shouldn't the null check occur _before_ wrapping the detector?

anomdetect/src/main/java/com/expedia/adaptivealerting/anomdetect/PerfMonListener.java
```diff
@@ -0,0 +1,30 @@
...
+ * monitor resets an evaluator
+ * 
+ * @author kashah
+ */
+public interface PerfMonListener {
+
+    /**
+     * Keeps a track of when performance monitor resets the evaluator.
+     */
+    public void readyForFlush(double score);
```
williewheeler: 1) public not necessary for interfaces. Interface methods are always public.
2) My `readyForFlush` suggestion was for the polling approach, not for the listener approach. For the listener a better name would be `processScore`.

anomdetect/src/main/java/com/expedia/adaptivealerting/anomdetect/PerformanceMonitor.java
```diff
@@ -34,37 +35,46 @@
...
+     */
+
+    private int maxTicks;
 
     /**
      * Creates a new performance monitor and sets evaluator as RMSE evaluator
      */
-    public PerformanceMonitor() {
+    public PerformanceMonitor(PerfMonListener listener, Evaluator evaluator, int maxTicks) {
+        this.listener = listener;
```
williewheeler: Please validate the params here, as this is an API method.. Neither listener nor evaluator should be allowed to be null, and maxTicks should be > 0.

anomdetect/src/main/java/com/expedia/adaptivealerting/anomdetect/MonitoredDetector.java
```diff
@@ -46,11 +46,10 @@ public MonitoredDetector(AnomalyDetector detector) {
 
     @Override
     public MappedMpoint classify(MappedMpoint mappedMpoint) {
-        MappedMpoint mappedPoint = detector.classify(mappedMpoint);
         PerformanceMonitor perfMonitor = new PerformanceMonitor(new PerfMonHandler(), new RmseEvaluator(), MAX_TICKS);
```
williewheeler: We don't want to create a new perfmon handler or RMSE evaluator on every classification. That could be thousands per second. Pass the perfmon into the monitored detector when you create that.

anomdetect/src/main/java/com/expedia/adaptivealerting/anomdetect/AnomalyDetectorManager.java
```diff
@@ -107,8 +107,9 @@ public AnomalyDetector detectorFor(MappedMpoint mappedMpoint) {
     public MappedMpoint classify(MappedMpoint mappedMpoint) {
         notNull(mappedMpoint, "mappedMpoint can't be null");
         final AnomalyDetector detector = detectorFor(mappedMpoint);
+        MappedMpoint newMappedMpoint = detector == null ? null : detector.classify(mappedMpoint);
```
williewheeler: 1) Don't pass null detector into MonitoredDetector. If the detector is null just return null from the method.
2) We don't want to create a new wrapper with every classification. The wrapper is what we store in the map.

PR #96, opened 2018-07-16
shewhart individuls chart implementation

shsethi:


anomdetect/src/main/java/com/expedia/adaptivealerting/anomdetect/ShewhartIndividuals.java
```diff
@@ -0,0 +1,185 @@
...
+        double currentRange = abs(prevValue - observed);
+        this.movingRangeSum += abs(currentRange);
+        this.runningMean = getRunningMean(observed);
+
+        final double dist = abs(observed - target);
+        this.totalDataPoints++;
+
+        AnomalyLevel anomalyLevel = UNKNOWN;
+        if (totalDataPoints <= warmUpPeriod || ((totalDataPoints-warmUpPeriod) % RECOMPUTE_LIMITS_PERIOD )==0) {
+
```
karanshah03: We should only start calculating control limits once the warm-up period value is reached.

anomdetect/src/main/java/com/expedia/adaptivealerting/anomdetect/ShewhartIndividuals.java
```diff
@@ -0,0 +1,185 @@
...
+            double averageMovingRange = getAverageMovingRange();
+            this.target = this.runningMean;
+
+            upperControlLimit_R = R_CONTROL_CHART_CONSTANT_D4 * averageMovingRange;
+            upperControlLimit_X = this.target + 2.66 * averageMovingRange;
+            lowerControlLimit_X = this.target - 2.66 * averageMovingRange;
+
+        } else {
+            anomalyLevel = NORMAL;
```
karanshah03: We may need to know it's left-tailed and right-tailed (user-defined param) and depending on that we calculate anomaly levels. left-tailed is for lower threshold and right-tailed is for upper threshold.

anomdetect/src/main/java/com/expedia/adaptivealerting/anomdetect/ShewhartIndividuals.java
```diff
@@ -0,0 +1,185 @@
...
+        result.setStrongThresholdUpper(upperControlLimit_X);
+        result.setWeakThresholdLower(lowerControlLimit_X);
+        result.setStrongThresholdLower(lowerControlLimit_X);
+        result.setAnomalyScore(dist);
+        result.setAnomalyLevel(anomalyLevel);
+        return result;
+    }
+
+    private double getRunningMean(double observed) {
+        return ( (this.runningMean * this.totalDataPoints) + observed) / (this.totalDataPoints + 1);
```
karanshah03: if the idea here is just to calculate mean?

PR #97, opened 2018-07-20
Perf monitor kafka

karanshah03:


williewheeler:
Karan, I've closed this as it is old and we need to rethink how we approach detector evaluation. RMSE is problematic in cases where the model simply tracks current observations (e.g. EWMA with alpha=1).

PR #101, opened 2018-07-29
model service schema and stored procs

shsethi:


modelservice/src/main/sql/build-db.sql
```diff
@@ -0,0 +1,74 @@
...
+USE model_service;
+
+
+CREATE TABLE metric
+(
+  metric_id   INT UNSIGNED PRIMARY KEY NOT NULL AUTO_INCREMENT,
+  metric_uuid CHAR(32) UNIQUE          NOT NULL
+);
+
+CREATE TABLE models
```
williewheeler: For consistency, we should either always use singular or always use plural. (So not "metric" for one table" and "models" for another.) My request would be that we just use singular for the table names.

modelservice/src/main/sql/build-db.sql
```diff
@@ -0,0 +1,74 @@
...
+  model_id      MEDIUMINT UNSIGNED PRIMARY KEY      NOT NULL AUTO_INCREMENT,
+  model_uuid    CHAR(32) UNIQUE                     NOT NULL,
+  hyperparams   JSON,
+  thresholds    JSON,
+  to_rebuild    BOOLEAN                                      DEFAULT FALSE,
+  last_build_ts TIMESTAMP
+);
+
+
+CREATE TABLE metric_model_map
```
williewheeler: Add a unique index (metric_id, model_id)

modelservice/src/main/sql/build-db.sql
```diff
@@ -0,0 +1,74 @@
...
+(
+  metric_id INT UNSIGNED       NOT NULL,
+  model_id  MEDIUMINT UNSIGNED NOT NULL
+);
```
williewheeler: Add a unique index (metric_id, model_id)

modelservice/src/main/sql/build-db.sql
```diff
@@ -0,0 +1,74 @@
...
+
+    INSERT INTO  models (model_uuid, hyperparams) VALUES (modelUuid, hyperParams);
+    INSERT INTO metric_model_map (metric_id,model_id) VALUES (metricID, LAST_INSERT_ID());
+
+  END;
+
+DROP PROCEDURE IF EXISTS update_threshold;
+CREATE PROCEDURE update_threshold(
+  metricUuid CHAR(32),
+  thresHold JSON,
```
williewheeler: This should be "threshold", not "thresHold".

modelservice/src/main/sql/build-db.sql
```diff
@@ -0,0 +1,74 @@
...
+
+-- =====================================================================================================================
+-- PROCEDURES
+-- =====================================================================================================================
+
+DROP PROCEDURE IF EXISTS add_new_model;
+
+CREATE PROCEDURE add_new_model(
+  metricUuid CHAR(32),
+  hyperParams JSON,
```
williewheeler: Again please be consistent with the names. In the models table above this is "hyperparams" but here is "hyperParams". It should be "hyperparams".

If the idea was to come up with a way to distinguish the proc params from the column names, I'd ask that we come up with some convention and use it throughout. For example we could use a _ prefix on the proc params in general.

In general it's probably better to use lowercase and _ throughout. That is pretty typical with MySQL, especially since MySQL is case sensitive sometimes and not case sensitive other times. It's too confusing to use mixed case in this sort of context. Please see http://mysql-tools.com/en/identifiers-in-mysql/15-identifiers-in-mysql.html:

"Note: Avoid mixing cases within the same statement:

mysql> SELECT * FROM Film WHERE FILM.film_id = 101;
Column names, column aliases, index names, stored routines and event names are not case sensitive on any platform.

However, names of triggers and logfile groups are case sensitive. This is a variation on standard SQL.

By default, table aliases are case sensitive on Unix, but not so on Windows or Mac OS X. The following statement would work on Windows but not Unix, because it refers to the alias in mixed case.

mysql> SELECT rating FROM film AS f WHERE f.film_id = 101 OR F.rental_rate = 4.99;
Tip: For convention sake, if you use mixed case to name your database objects, it is best to use that convention throughout. Another common option is use of underscores to separate elements in a name, as in car_make.

To avoid such problems, for maximum portability and ease of use, it is best to be consistent conventionally, such as always creating and referring to databases and tables using lowercase names."

PR #134, opened 2018-08-07
[Model Service] modifying jar packaging to run in Docker

shsethi:
To verify do following

```
mvn clean package

make docker_build

docker run haystack-adaptive-alerting-modelservice:latest

```


modelservice/src/main/resources/log4j.properties
```diff
@@ -0,0 +1,27 @@
...
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+#
+
+# Root logger option
+log4j.rootLogger=INFO, stdout
+
+# Packages and classes
+log4j.logger.com.expedia.adaptivealerting.modelservice=TRACE, stdout
```
williewheeler: This will double-log. Either remove stdout from this line or else set additivity to false.
https://stackoverflow.com/questions/5699309/log4j-logging-twice

modelservice/src/main/resources/log4j.properties
```diff
@@ -0,0 +1,28 @@
...
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+#
+
+# Root logger option
+log4j.rootLogger=INFO, stdout
+log4j.additivity.org.hibernate=false
```
williewheeler: Additivity would need to be for `com.expedia.adaptivealerting.modelservice` rather than Hibernate.

PR #136, opened 2018-08-09
Enhancement/model service iam auth

karanshah03:


modelservice/src/main/java/com/expedia/adaptivealerting/modelservice/util/ThreadUtil.java
```diff
@@ -0,0 +1,25 @@
...
+ */
+package com.expedia.adaptivealerting.modelservice.util;
+
+import java.util.concurrent.TimeUnit;
+
+public class ThreadUtil {
+
+    public static void sleep(int mins) throws InterruptedException {
+        TimeUnit.MINUTES.sleep(mins);
+    }
```
williewheeler: may as well put the RuntimeException in here, no?

modelservice/src/main/resources/config/application.yml
```diff
@@ -2,14 +2,10 @@ spring:
...
     show-banner: true
-  datasource:
-    driver-class-name: com.mysql.jdbc.Driver
-    url: jdbc:mysql://127.0.0.1:3306/model_service?autoReconnect=true
-    username: root
-    password:
-    maximum-pool-size: 8
-    minimum-idle: 0
-    # TODO Add other data source settings here
+    datasource:
```
williewheeler: This goes under spring, not main

modelservice/src/main/resources/config/application.yml
```diff
@@ -27,6 +23,8 @@ management:
...
 server:
   port: 8080
- 
- 
- 
\ No newline at end of file
+  
+db:
+  url: "jdbc:mysql://aa-model-db.c24oooc4uqmr.us-west-2.rds.amazonaws.com:3306/model_service?autoReconnect=true"
+  user: "aa_user"
```
williewheeler: ?? this is expedia-specific. are you just trying to get it to work? if so that's fine, but we need to remove it and not forget to do so

PR #145, opened 2018-08-16
Enhancement/swagger documentation

karanshah03:


modelservice/src/main/java/com/expedia/adaptivealerting/modelservice/web/ModelController.java
```diff
@@ -41,29 +45,34 @@
     public Boolean getModel() {
         return true;
     }
+
+    @ApiOperation(value = "Get model info by providing metric key")
     @RequestMapping(value = "/api/model", method = RequestMethod.GET, produces = MediaType.APPLICATION_JSON_VALUE)
```
williewheeler: Move `/api` up to the class level. Also I would prefer to leave the `produces` element off here--if the client accepts `text/xml` then it would be weird to force a json response. What happens if you remove this? Do we try to return XML? (We don't want to return XML but I'd prefer to 406 instead of returning JSON.)

modelservice/src/main/java/com/expedia/adaptivealerting/modelservice/web/ModelController.java
```diff
@@ -41,29 +45,34 @@
...
-    @RequestMapping(value = "/api/markToRebuild", method = RequestMethod.PUT)
-    public HttpStatus markToRebuild(@RequestBody RebuildParams rebuildParams) {
+    @ApiOperation(value = "Mark model to rebuild")
+    @RequestMapping(value = "/api/markToRebuild", method = RequestMethod.PUT, produces = MediaType.APPLICATION_JSON_VALUE)
+    public ResponseEntity<String> markToRebuild(@RequestBody RebuildParams rebuildParams) {
         modelService.markToRebuild(rebuildParams.getModelUUID(), rebuildParams.getMetricKey(),
                 rebuildParams.getToRebuild());
-        return HttpStatus.OK;
+        return new ResponseEntity<String>("Model marked for rebuild", HttpStatus.OK);
     }
```
williewheeler: Simpler: return a String, and just return "Model marked for rebuild" as the string. Spring will automatically 200 the response. Same for the other cases. Keep the code tight here.

modelservice/src/main/java/com/expedia/adaptivealerting/modelservice/web/ModelController.java
williewheeler: Also if you do what I'm saying, you can use `@GetMapping`.

modelservice/pom.xml
```diff
@@ -83,6 +83,16 @@
...
+        <!-- Swagger -->
+        <dependency>
+            <groupId>io.springfox</groupId>
+            <artifactId>springfox-swagger-ui</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.springfox</groupId>
+            <artifactId>springfox-swagger2</artifactId>
+        </dependency>
+
```
williewheeler: Please use the commit/runtime/test scheme for grouping the dependencies. (We were originally using the library names but decided to go with the type.)

modelservice/src/main/java/com/expedia/adaptivealerting/modelservice/web/WebController.java
```diff
@@ -0,0 +1,32 @@
...
+
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.RestController;
+
+/**
+ * @author kashah
+ *
+ */
+@RestController
+public class WebController {
```
williewheeler: This is a poor name as all controllers here will be web controllers. Choose a name that clarifies the responsibility this web controller has relative to others. For example this could be `HealthController` or similar.

pom.xml
```diff
@@ -312,6 +313,20 @@
...
+                <version>${swagger.version}</version>
+                <scope>compile</scope>
+            </dependency>
+            <dependency>
+                <groupId>io.springfox</groupId>
+                <artifactId>springfox-swagger2</artifactId>
+                <version>${swagger.version}</version>
+                <scope>compile</scope>
+            </dependency>
+
```
williewheeler: See comments above. Move this to the compile section and preserve alphabetical order so we don't end up with duplicates.

PR #149, opened 2018-09-03
Bug fix/model service

karanshah03:


modelservice/src/main/java/com/expedia/adaptivealerting/modelservice/service/impl/ModelServiceImpl.java
```diff
@@ -42,26 +44,56 @@
...
+            metric.getModels().add(model);
+            modelRepository.save(model);
+            return "Model params saved successfully";
+        } else {
+            return "Metric-Model mapping already exists";
+        }
     }
 
     @Override
     public void markToRebuild(String modelUUID, String metricKey, Boolean toRebuild) {
```
williewheeler: Why does this allow a Boolean instead of requiring a boolean?

modelservice/src/main/java/com/expedia/adaptivealerting/modelservice/service/impl/ModelServiceImpl.java
```diff
@@ -42,26 +44,56 @@
...
+        }
+
+        if (metricId == null || modelId == null) {
+            model.getMetrics().add(metric);
+            metric.getModels().add(model);
+            modelRepository.save(model);
+            return "Model params saved successfully";
+        } else {
+            return "Metric-Model mapping already exists";
+        }
```
williewheeler: These return values are strange for a model service. Normally we would not use display text as a return value for a successful operation as human-friendly display text is a UI concern, not a service concern. I would expect to see void return type here and an exception if anything goes wrong.

modelservice/src/main/java/com/expedia/adaptivealerting/modelservice/service/impl/ModelServiceImpl.java
```diff
@@ -42,26 +44,56 @@
...
     @Override
-    public void addModelParams(ModelParams modelParams) {
+    public String addModelParams(ModelParams modelParams) {
 
-        Model model = new Model(modelParams.getModelUUID(), modelParams.getHyperparams());
-        Metric metric = new Metric(modelParams.getMetricKey());
-        model.getMetrics().add(metric);
-        metric.getModels().add(model);
-        modelRepository.save(model);
+        AssertUtil.notNull(modelParams.getHyperparams(), "Hyper params can't be null");
```
williewheeler: "hyperparams", not "hyper params"

modelservice/src/main/java/com/expedia/adaptivealerting/modelservice/web/ModelController.java
```diff
@@ -46,8 +46,7 @@
     @ApiOperation(value = "Add model hyper params")
```
williewheeler: "hyperparams", not "hyper params"

modelservice/src/main/java/com/expedia/adaptivealerting/modelservice/web/ModelController.java
```diff
@@ -46,8 +46,7 @@
...
     @PostMapping(value = "/addModelParams", produces = MediaType.APPLICATION_JSON_VALUE)
```
williewheeler: The REST URIs should generally be resources, not operations. I would expect to see something like

/models/:uuid

and then do an HTTP PUT against that. Are we really _adding_ params here, or just setting them? (The params _are_ the model...)

modelservice/src/main/java/com/expedia/adaptivealerting/modelservice/service/impl/ModelServiceImpl.java
williewheeler: Also I would not expect the model params to have a reference to hyperparameters

modelservice/src/main/java/com/expedia/adaptivealerting/modelservice/dto/Hyperparams.java
```diff
@@ -24,9 +24,8 @@
  */
 
 @Data
-public class ModelParams {
+public class Hyperparams {
 
-    private String modelUUID;
     private String metricKey;
```
williewheeler: Why are we binding to a specific metric here? Does the hyperparams object need to know about the metric at all? In fact does it need to know about the model?

I wonder--do we need a Hyperparams class at all? Won't it simply store a dictionary that maps a hyperparam to a hyperparam value? If so, could we just use a Properties in the model itself? (I.e. what would Hyperparams class be doing other than wrapping a Properties?)

modelservice/src/main/java/com/expedia/adaptivealerting/modelservice/service/ModelService.java
```diff
@@ -28,9 +28,9 @@
 
     List<ModelDto> getModels(String metricKey);
 
-    void addModelParams(ModelParams modelParams);
+    void addModelParams(String modelUUID, Hyperparams hyperparams);
```
williewheeler: Is `addModelParams` the right name for this method? (Please see the comments from the last code review around this point.)

modelservice/src/main/java/com/expedia/adaptivealerting/modelservice/service/ModelService.java
```diff
@@ -28,9 +28,9 @@
...
 
-    void markToRebuild(String modelUUID, String metricKey, Boolean toRebuild);
+    void markToRebuild(String modelUUID, String metricKey, boolean toRebuild);
 
     void updateThresholds(String modelUUID, String metricKey, Map<String, Object> thresholds);
```
williewheeler: I believe there are only two thresholds here: `weakSigmas` and `strongSigmas`. (I like "sigmas" here because it makes it explicit what sort of threshold we are talking about--it's not the constant thresholds.) If this is right then we don't want a general string-to-object map.

modelservice/src/main/java/com/expedia/adaptivealerting/modelservice/web/ModelController.java
```diff
@@ -43,25 +44,26 @@
...
+        modelService.addModelParams(uuid, hyperparams);
         return "Model params saved successfully";
     }
 
     @ApiOperation(value = "Mark model to rebuild")
-    @PutMapping(value = "/markToRebuild", produces = MediaType.APPLICATION_JSON_VALUE)
-    public String markToRebuild(@RequestBody RebuildParams rebuildParams) {
-        modelService.markToRebuild(rebuildParams.getModelUUID(), rebuildParams.getMetricKey(),
-                rebuildParams.getToRebuild());
+    @PutMapping(value = "/rebuildModel/{uuid}", produces = MediaType.APPLICATION_JSON_VALUE)
```
williewheeler: REST URIs should be nouns, not verbs. (Use `/models/:uuid` rather than `/rebuildModel/:uuid`. See previous PR for more info.

Also we need to store the rebuild schedule, right? E.g. rebuild every 24 hours or whatever. Would it be better to store the last scheduled rebuild time and the rebuild period, and then do schedule a rebuild whenever now <= lastScheduled + period? That way we don't require some external actor to come in and set this `toRebuild` flag.

modelservice/src/main/java/com/expedia/adaptivealerting/modelservice/web/ModelController.java
```diff
@@ -43,25 +44,26 @@
...
+                rebuildParams.isToRebuild());
         return "Model marked for rebuild";
     }
 
+    @Deprecated
     @ApiOperation(value = "Update threshold for a given model")
-    @PutMapping(value = "/updateThreshold", produces = MediaType.APPLICATION_JSON_VALUE)
-    public String updateThresholds(@RequestBody ThresholdParams thresholdParams) {
-        modelService.updateThresholds(thresholdParams.getModelUUID(), thresholdParams.getMetricKey(),
+    @PutMapping(value = "/updateThreshold/{uuid}", produces = MediaType.APPLICATION_JSON_VALUE)
```
williewheeler: REST endpoints should be nouns. Try

`/models/:uuid/thresholds`

or something similar

PR #155, opened 2018-09-12
bug fix to allow many to many mapping b/w metric and model

karanshah03:


modelservice/src/main/java/com/expedia/adaptivealerting/modelservice/repo/MetricRepository.java
```diff
@@ -17,12 +17,14 @@
...
 
 /**
  * @author kashah
- *
  */
 public interface MetricRepository extends JpaRepository<Metric, Integer> {
 
-    Metric findMetricByMetricKey(String metricKey);
+    @Query("select m.id from Metric m where m.metricKey = :key")
+    Integer findIdByMetricKey(@Param("key") String key);
```
williewheeler: Wouldn't we want to find the metric by the metric _hash_ here? Not the key. Not all metrics have a key.

Instead of finding the ID by hash, just find the whole Metric by hash using `findByHash`. This is more generally useful and doesn't change the query performance at all.

But we are not going to want to make two separate queries to get the models for a given metric--we can do it with a single query.

One constraint on the query is that we don't want metrics to know about models and vice versa. These are independent concepts. So what that means is that the `metric_model` link table should actually become its own entity in the Java representation. (Consider renaming the table to `metric_model_mapping` as well.) Let me suggest creating a `MetricModelMapping` entity along with a `MetricModelMappingRepo` repo. The `MetricModelMapping` should have a single reference to a `Metric` and another reference to a `Model`. Then create a JPA method on the repo called `findByMetricHash(String)`. This should automatically pull up all desired mappings in a single query. Moreover you will not need to define an explicit `@Query`.

modelservice/src/main/java/com/expedia/adaptivealerting/modelservice/repo/ModelRepository.java
```diff
@@ -17,13 +17,16 @@
...
 
 /**
  * @author kashah
  *
  */
 public interface ModelRepository extends JpaRepository<Model, Integer> {
 
-    Model getModelById(Integer id);
+    @Query("select m.id from Model m where m.modelUUID = :uuid")
+    Integer findIdByModelUUID(@Param("uuid") String uuid);
```
williewheeler: Similar comments apply here. I don't see any advantage to projecting the query results to a single ID. Why not simply return the model itself? Same query behind the scenes.

modelservice/src/main/java/com/expedia/adaptivealerting/modelservice/service/ModelService.java
```diff
@@ -28,9 +28,9 @@
 
     List<ModelDto> getModels(String metricKey);
 
-    void addModelParams(ModelParams modelParams);
+    void addHyperparams(String modelUUID, Hyperparams hyperparams);
```
williewheeler: Can you explain why we are "adding" hyperparams here? Why wouldn't we just save the whole model with its hyperparams as a single action? And anyway, even if we were to do this as a two actions for whatever reason, we would be _setting_ the hyperparameters, not _adding_ them. Adding connotes augmenting a collection. Here we are setting the hyperparams on the model. But please prefer just saving the model with its hyperparams in one action unless there is some reason to do it as two actions.

modelservice/src/main/java/com/expedia/adaptivealerting/modelservice/service/impl/ModelServiceImpl.java
```diff
@@ -42,26 +44,55 @@
...
+    private MetricRepository metricRepository;
+
     @Override
     public List<ModelDto> getModels(String metricKey) {
         return modelRepositoryCustom.findModels(metricKey);
     }
 
     @Override
-    public void addModelParams(ModelParams modelParams) {
+    public void addHyperparams(String uuid, Hyperparams hyperparams) {
```
williewheeler: See comments above.

modelservice/src/main/java/com/expedia/adaptivealerting/modelservice/service/impl/ModelServiceImpl.java
```diff
@@ -42,26 +44,55 @@
...
+            metric.getModels().add(model);
+            modelRepository.save(model);
+        } else {
+            throw new RuntimeException("Metric-Model mapping already exits");
+        }
     }
 
     @Override
-    public void markToRebuild(String modelUUID, String metricKey, Boolean toRebuild) {
+    public void markToRebuild(String modelUUID, String metricKey, boolean toRebuild) {
```
williewheeler: Let's chat about this one in person. I am not convinced that we want some external agent to mark these models for rebuild, at least at this time. Presumably the rebuild triggers would either be cron-based or else performance-based.

For a cron-based approach I think we'd want to trigger off of a declared schedule, and we may as well put that schedule in the model service instead of assuming some external thing know the schedule and call `ModelService.markToRebuild()`. (No reason to push the problem back.)

For the performance-based approach, we would need some good way to measure model performance, and we don't yet have that. (I don't any longer think RMSE is the right thing here--that's prediction model performance, not classification model performance, and it's trivial to make the predictions _perfect_ and have bad classifications.)

So my suggestion is to get rid of `markToRebuild` for now, and let's figure out how to define a rebuild schedule.

modelservice/src/main/java/com/expedia/adaptivealerting/modelservice/service/impl/ModelServiceImpl.java
```diff
@@ -70,9 +101,12 @@ public void markToRebuild(String modelUUID, String metricKey, Boolean toRebuild)
     @Override
     public void updateThresholds(String modelUUID, String metricKey, Map<String, Object> thresholds) {
 
+        AssertUtil.notNull(modelUUID, "Model UUID can't be null");
+        AssertUtil.notNull(thresholds, "Thresholds can't be null");
+
         Integer modelID = modelRepositoryCustom.getModelID(metricKey, modelUUID);
-        Model model = modelRepository.getModelById(modelID);
+        Model model = modelRepository.getOne(modelID);
```
williewheeler: Ha, this is a great example of where my suggestion of just returning the `Model` would allow you to have a single call instead of two.

Also, you shouldn't be calling `modelRepositoryCustom`. All repo calls should go through the `ModelRepository` abstraction. You can wire custom implementation code into `ModelRepository` interface methods. See the Spring Data JPA docs.

modelservice/src/test/java/com/expedia/adaptivealerting/modelservice/web/HealhControllerTests.java
```diff
@@ -0,0 +1,27 @@
+package com.expedia.adaptivealerting.modelservice.web;
+
+import org.junit.Before;
+import org.junit.Test;
+import org.mockito.InjectMocks;
+import org.mockito.MockitoAnnotations;
+
+import static junit.framework.TestCase.assertEquals;
+
+public class HealhControllerTests {
```
williewheeler: HealthControllerTests

PR #166, opened 2018-09-19
Enhancement/model service

karanshah03:


modelservice/src/main/sql/build-db.sql
```diff
@@ -7,10 +7,11 @@ use aa_model_service;
 
 create table metric
 (
-  id        int unsigned primary key not null auto_increment,
-  m_key     varchar(100) unique not null,
-  hash      varchar(100) unique not null,
-  tags      json
+  id           int unsigned primary key not null auto_increment,
+  m_key        varchar(100) unique not null,
```
williewheeler: Use `key` (backticks) instead of m_key please. One of the other keys uses `key` so let's be consistent.

modelservice/src/main/sql/build-db.sql
```diff
@@ -7,10 +7,11 @@ use aa_model_service;
...
+  hash         varchar(100) unique not null,
```
williewheeler: Use char(n) if the hash has a fixed length, which I suspect it does.

modelservice/src/main/sql/build-db.sql
```diff
@@ -7,10 +7,11 @@ use aa_model_service;
...
+  description  varchar(100) unique not null,
```
williewheeler: description should probably be longer than 100. I could imagine somebody wanting to describe the data source, or putting a link to a doppler URL, etc. can we do 255 here?

williewheeler:
@karanshah03 Build failed but looks environmental. Will take a look.

modelservice/src/main/java/com/expedia/adaptivealerting/modelservice/repo/ModelRepository.java
```diff
@@ -26,6 +27,10 @@
  * @author kashah
  */
 public interface ModelRepository extends PagingAndSortingRepository<Model, Long> {
+
     List<Model> findByUuid(@Param("uuid") String uuid);
 
+    @Query("select mmm.model from MetricModelMapping mmm where mmm.metric.hash=:hash")
+    List<Model> findModelsByHash(@Param("hash") String hash);
```
williewheeler: Let's call this "findByMetricHash". The "findModels" part is implicit since this is on the model repo.

modelservice/src/main/sql/build-db.sql
```diff
@@ -7,16 +7,18 @@ use aa_model_service;
...
+  tags         json
 );
 
 create table model_type
 (
-  id        smallint unsigned primary key not null auto_increment,
-  `key`     varchar(100) unique not null
+  id           smallint unsigned primary key not null auto_increment,
+  `key`        varchar(100) unique not null,
+  seyren_flag  boolean default false
```
williewheeler: This belongs on the model table, not the model type table. It's individual models that we're trying to track as either coming from Seyren or not.

modelservice/src/main/java/com/expedia/adaptivealerting/modelservice/entity/ModelType.java
```diff
@@ -30,10 +30,13 @@
...
+
     @Id
     @GeneratedValue(strategy = GenerationType.IDENTITY)
     private Long id;
-    
-    public String key;
+
+    private String key;
+
+    private boolean seyrenFlag;
```
williewheeler: 1. This belongs on the model table, not the model_type table. See below.
2. Let's not expose this on the entity itself. This is a purely internal flag for now and we don't want people to start developing to it.

williewheeler:
Submitting new PR, stay tuned.

PR #207, opened 2018-10-01
changing application id of validator

shsethi:


shsethi:
Change needs to be in k8s

PR #211, opened 2018-10-05
AA notification webhook

bibinss:


dataconnect/pom.xml
```diff
@@ -46,7 +46,7 @@
         <!-- TODO Currently this is a direct dependency. Soon we will replace this with a web service call. [WLW] -->
         <dependency>
             <groupId>com.expedia.adaptivealerting</groupId>
-            <artifactId>adaptive-alerting-dataservice-s3</artifactId>
```
bibinss: local build was failing due to this stale dependency

notifier/src/main/java/com/expedia/adaptivealerting/notifier/service/Notifier.java
```diff
@@ -0,0 +1,107 @@
...
+            consumerRecords.forEach( record -> {
+                buildJson(record.value()).ifPresent(json -> {
+                    HttpHeaders headers = new HttpHeaders();
+                    headers.setContentType(MediaType.APPLICATION_JSON);
+                    HttpEntity<String> entity = new HttpEntity<>(json, headers);
+                    try {
+                        ResponseEntity responseEntity =
+                            restTemplate.exchange(webhookUrl, HttpMethod.POST, entity, ResponseEntity.class);
+                        if (responseEntity.getStatusCode().is2xxSuccessful()) {
+                            MetricsMonitor.notification_success.mark();
```
bibinss: meter to track failures and success. i think probably we should add counters/meters in other AA apps to track the data flow, success and failures.

williewheeler: Good call

PR #219, opened 2018-10-08
Update stored-procs.sql

dinilatgit:
Adding the stored procs for wildcard metrics mapping to model.
During the migration, the mappings between model and wildcard metric ukey will be added to map_wildcard_metrics_to_model

### DIFF
--- a/modelservice/src/main/sql/stored-procs.sql
+++ b/modelservice/src/main/sql/stored-procs.sql
@@ -2,6 +2,9 @@ USE `aa_model_service`;
 
 DROP PROCEDURE IF EXISTS insert_mapping;
 DROP PROCEDURE IF EXISTS insert_model;
+DROP PROCEDURE IF EXISTS insert_mapping_multiple_metrics_to_model;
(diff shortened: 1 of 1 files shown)


williewheeler:
I’m not clear that this should be part of the AA project. Do we see this as something that we would use outside the Seyren migration? Also I don’t think we want to use these scripts as a way to  manage lists of expedia-internal metrics.

dinilatgit:
> I’m not clear that this should be part of the AA project. Do we see this as something that we would use outside the Seyren migration? Also I don’t think we want to use these scripts as a way to manage lists of expedia-internal metrics.

The procedure that calls the mapping proc to update could be used as a migration script hence being discarded from the change. Retaining the mapping proc alone as a utility.
New change - https://github.com/ExpediaDotCom/adaptive-alerting/pull/221

PRs https://github.com/ExpediaDotCom/adaptive-alerting/pull/219, https://github.com/ExpediaDotCom/adaptive-alerting/pull/220 are marked closed.

PR #222, opened 2018-10-09
Updates build so that it works on JDK 11

### LINKED ISSUE #217: make things work with JDK 11
We've done some things in zipkin to make things build with JDK 10+ I can help with this.

adriancole:
Notable change is ensuring more recent scala is used, which implies
Kafka 1.1.1+

Also, a couple tests were never run due to naming convention. This
renames them such that they run.

Fixes #217

### DIFF
--- a/kafka/pom.xml
+++ b/kafka/pom.xml
@@ -34,9 +34,10 @@
     <description>Kafka interface for Adaptive Alerting.</description>
 
     <properties>
-        <maven.scalatest.plugin.version>1.0</maven.scalatest.plugin.version>
-        <maven.scala.plugin.version>3.2.1</maven.scala.plugin.version>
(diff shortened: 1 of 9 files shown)


notifier/pom.xml
```diff
@@ -32,11 +32,6 @@
             <groupId>com.expedia.adaptivealerting</groupId>
             <artifactId>adaptive-alerting-kafka</artifactId>
         </dependency>
-        <dependency>
```
adriancole: redundant

kafka/pom.xml
```diff
@@ -128,26 +117,9 @@
     <build>
         <testSourceDirectory>src/test/scala</testSourceDirectory>
         <plugins>
-            <plugin>
```
adriancole: removed unused things in this file

williewheeler:
Thanks @adriancole!

PR #231, opened 2018-10-17
[WIP] read models from modelservice db instead of S3

shsethi:


adriancole:
cool in concept. haven't looked at code

anomdetect/src/main/java/com/expedia/adaptivealerting/anomdetect/BasicAnomalyDetectorFactory.java
```diff
@@ -42,41 +43,32 @@
...
+
+        final Resources<ModelResource> modelResources = modelServiceConnector.findModels(uuid);
+        final Collection<ModelResource> modelResourceCollection = modelResources.getContent();
+
+        List<ModelResource> modelResourceList = new ArrayList<>(modelResourceCollection);
+
+        if (modelResourceList.size() < 1) {
+            log.error("No model found for uuid = {}", uuid);
+        }
+        final AnomalyDetectorModel model = modelResourceList.get(0);
```
williewheeler: Compare this to the previous block where you check for modelResourceList.size() < 1. If that condition obtains then you will get an IndexOutOfBoundsException here. Decide how you want to handle the case where modelResourceList.size() < 1 and then make sure the code supports it.

PR #232, opened 2018-10-18
read models from modelservice db instead of S3,  Implement basic detector

djsutho:
Just building off #231 just a quick fix for now around getting the `BasicAnomalyDetectorFactory` working for our basic detectors. We will want to clean this up in future.

### DIFF
--- a/anomdetect/src/main/java/com/expedia/adaptivealerting/anomdetect/BasicAnomalyDetectorFactory.java
+++ b/anomdetect/src/main/java/com/expedia/adaptivealerting/anomdetect/BasicAnomalyDetectorFactory.java
@@ -15,16 +15,17 @@
  */
 package com.expedia.adaptivealerting.anomdetect;
 
-import com.amazonaws.services.s3.AmazonS3;
-import com.amazonaws.services.s3.AmazonS3ClientBuilder;
-import com.amazonaws.services.s3.model.S3Object;
+import com.expedia.adaptivealerting.anomdetect.util.ModelResource;
+import com.expedia.adaptivealerting.anomdetect.util.ModelServiceConnector;
 import com.expedia.adaptivealerting.core.util.ReflectionUtil;
-import com.expedia.adaptivealerting.core.util.jackson.ObjectMapperUtil;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.typesafe.config.Config;
 import lombok.extern.slf4j.Slf4j;
+import org.springframework.hateoas.Resources;
 
-import java.io.InputStream;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
 import java.util.UUID;
 
 import static com.expedia.adaptivealerting.core.util.AssertUtil.notNull;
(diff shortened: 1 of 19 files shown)


anomdetect/src/main/java/com/expedia/adaptivealerting/anomdetect/constant/ConstantThresholdModel.java
```diff
@@ -26,6 +27,6 @@
 @JsonIgnoreProperties(ignoreUnknown = true)
 public class ConstantThresholdModel implements AnomalyDetectorModel {
     private String uuid;
-    private String type;
+    private ModelTypeResource detectorType;
```
williewheeler: `AnomalyDetectorModel` implementations should not have a `ModelTypeResource`. The implementation's class _is_ the type. In the database we need the type as a field since we're using a single model table to handle all types (i.e. it's one of the three standard strategies for modeling inheritance in a database), but the mapping to Java should resolve that "impedance mismatch".

anomdetect/src/main/java/com/expedia/adaptivealerting/anomdetect/cusum/CusumAnomalyDetector.java
```diff
@@ -39,14 +39,11 @@
  * @author kashah
  */
 @Data
-public final class CusumAnomalyDetector implements AnomalyDetector {
+public final class CusumAnomalyDetector extends BasicAnomalyDetector<CusumParams> {
```
williewheeler: I wouldn't expect model internals (i.e. `CusumParams`) to make it all the way up to the detector generic params, just based on [Law of Demeter](https://en.wikipedia.org/wiki/Law_of_Demeter) considerations. Detector can know about models but model params are too deeper than what the detector should know or care about. I think what we want here is for the model to appear as a generic param, and push all the param stuff down into the actual model.

anomdetect/src/main/java/com/expedia/adaptivealerting/anomdetect/ewma/EwmaAnomalyDetector.java
```diff
@@ -47,11 +47,8 @@
  * @see <a href="https://www.itl.nist.gov/div898/handbook/pmc/section3/pmc324.htm">EWMA Control Charts</a>
  */
 @Data
-public final class EwmaAnomalyDetector implements AnomalyDetector {
-    
-    @NonNull
-    private UUID uuid;
-    
+public final class EwmaAnomalyDetector extends BasicAnomalyDetector<EwmaParams> {
```
williewheeler: Similar comments as above regarding detector/params and Law of Demeter. I'll stop mentioning it but take care of any cases.

kafka/src/main/resources/config/base.conf
```diff
@@ -80,6 +68,7 @@ ad-manager {
   }
   inbound-topic = "mapped-metrics"
   outbound-topic = "anomalies"
+  model-service-uri-template = "http://aa.local:32300/modelservice/api/models/search/findTopByDetectorUuidOrderByBuildTimestampDesc?uuid=%s"
```
williewheeler: We need to fix this long search name using `@Query`. Oh I see below that you are using `@Query` already but not renaming it. I suggest `findLatestByDetectorUuid`?

anomdetect/src/main/java/com/expedia/adaptivealerting/anomdetect/BasicAnomalyDetectorFactory.java
```diff
@@ -42,41 +43,33 @@
...
-        final S3Object s3Obj = s3.getObject(bucket, path);
-        final InputStream is = s3Obj.getObjectContent();
-        final AnomalyDetectorModel model = ObjectMapperUtil.readValue(objectMapper, is, AnomalyDetectorModel.class);
-        log.info("Loaded model: {}", model);
-        
+
+        final Resources<ModelResource> modelResources = modelServiceConnector.findModels(uuid);
+        final Collection<ModelResource> modelResourceCollection = modelResources.getContent();
+
+        List<ModelResource> modelResourceList = new ArrayList<>(modelResourceCollection);
```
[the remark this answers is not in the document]
williewheeler: @shsethi I notice that this is my mistake from a previous commit, not yours--you are just building onto what I already did. I am fixing it.

kafka/src/main/resources/config/base.conf
williewheeler: Fixed in PR #240

PR #237, opened 2018-10-24
Fix #235: Removed anomvalidate module

### LINKED ISSUE #235: Remove anomvalidate module
We've decided to limit AA's scope to time series anomaly detection and have validation occur as a separate step in the pipeline, outside of AA. Therefore we need to remove the anomvalidate module.

klymenkoo:
In this PR, I have removed anomvalidate module from the project as it was decided that it's going to be in a different step in pipeline as described in issue #235 

After removing the module, the project was built successfully, and all tests passed.


williewheeler:
Thanks @klymenkoo. Appreciate the assist. I'll work on getting some more issues posted in case you're interested in picking more up. Let me know if you have any particular areas of interest (e.g., machine learning, Kafka, etc.).

PR #238, opened 2018-10-26
Refactor/pr 232 follow up

williewheeler:


williewheeler:
I improved the model service search query names in PR #240

PR #239, opened 2018-10-26
added new endpoints to find metric and detector by owner

karanshah03:


modelservice/src/main/java/com/expedia/adaptivealerting/modelservice/repo/DetectorRepository.java
```diff
@@ -30,8 +30,28 @@
...
+     */
     List<Detector> findByUuid(@Param("uuid") String uuid);
 
+    /**
+     * Finds a list of detectors owned by provided owner, if any.
+     *
+     * @param owner Detector owner.
+     * @return List of detectors
+     */
+    List<Detector> findByCreatedBy(@Param("owner") String owner);
```
williewheeler: Let's use param "user" instead of "owner", since the createdBy column isn't necessarily the one who owns it. Fine if you want to treat the createdBy user as the owner from the UI, at least for now, but I don't want to conflate those two things in the API itself.

modelservice/src/main/java/com/expedia/adaptivealerting/modelservice/repo/MetricRepository.java
```diff
@@ -16,30 +16,43 @@
...
+
+
+    /**
+     * Finds a list of metrics by its owner.
+     *
+     * @param owner Detector owner.
+     * @return List of metrics by its owner
+     */
+    @Query(nativeQuery = true, value = "SELECT * FROM metric WHERE id IN (SELECT metric_id FROM metric_detector_mapping WHERE detector_id IN (SELECT id FROM detector WHERE created_by=:owner))")
+    List<Metric> findByDetectorOwner(@Param("owner") String owner);
```
williewheeler: Use "user" here please.

williewheeler: Please use findByCreatedBy() and param=user. Same reason as above.

Issue #218, opened 2018-10-08
Remove dependency on haystack-commons


williewheeler:
AA shouldn't have any Haystack dependencies. In particular we will need to ensure that we can handle the following without haystack-commons:

- Metric-related data structures and serdes
- KStream app configuration

williewheeler:
This is dev-complete, but I'm waiting til we complete our October deliverable before I push this since it's potentially disruptive.

williewheeler:
Completed in daeb5ef8891c492756d42732066813280fcd732f.

PR #257, opened 2018-11-02
Removes haystack-commons dependency per #218

williewheeler:


kafka/src/test/java/com.expedia.adaptivealerting.kafka/KafkaAnomalyDetectorManagerTest.java
```diff
@@ -0,0 +1,33 @@
...
+import static org.junit.Assert.fail;
+
+/**
+ * @author Willie Wheeler
+ */
+public class KafkaAnomalyDetectorManagerTest {
+    
+    @Test
+    @Ignore
+    public void testDummy() {
```
djsutho: Do we want to keep this? Is it a placeholder?

kafka/src/test/resources/config/base.conf
```diff
@@ -51,27 +49,19 @@ ad-manager {
...
-    default.timestamp.extractor = "com.expedia.adaptivealerting.kafka.serde.MappedMetricDataTimestampExtractor"
-    default.deserialization.exception.handler = "org.apache.kafka.streams.errors.LogAndContinueExceptionHandler"
+    acks = "all"
+    retries = "0"
+    key.serializer = "org.apache.kafka.common.serialization.StringSerializer"
+    value.deserializer = "com.expedia.adaptivealerting.kafka.serde.JsonPojoDeserializer"
   }
-  inbound-topic = "anomalies"
-  outbound-topic = "alerts"
+  topic = "constant-metrics"
```
djsutho: This looks outdated

pom.xml
```diff
@@ -541,6 +532,20 @@
...
+                            <goals>
+                                <goal>integration-test</goal>
+                                <goal>verify</goal>
+                            </goals>
+                            <configuration>
+                                <!-- TODO -->
+                            </configuration>
+                        </execution>
+                    </executions>
+                </plugin>                <plugin>
```
djsutho: `<plugin>` on new line

kafka/src/main/java/com/expedia/adaptivealerting/kafka/StreamsAppConfig.java
```diff
@@ -0,0 +1,83 @@
...
+    
+    /**
+     * Outbound topic configuration key.
+     */
+    private static final String CK_OUTBOUND_TOPIC = "outbound-topic";
+    
+    /**
+     * Health status path configuration key.
+     */
+    private static final String CK_HEALTH_STATUS_PATH = "health.status.path";
```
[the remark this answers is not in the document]
williewheeler: I would say let's leave it for now, because there is a good chance that I will reinstate the same health check mechanism after i get the deserialization exception handler working. I would rather just leave it there now as opposed to deleting it from the app and the terraform/k8s configs, only to re-add them. Having said that I will be sure to either use them or delete them.

kafka/src/test/java/com.expedia.adaptivealerting.kafka/KafkaAnomalyDetectorManagerTest.java
williewheeler: Yeah, just a placeholder to avoid JUnit complaining about no tests.

kafka/src/test/resources/config/base.conf
williewheeler: +1. I will move the whole config dir to the scala integration tests since that's what's intended. the current e2e test is ignored right now and needs a revamp. i will create a ticket for that

williewheeler: https://github.com/ExpediaDotCom/adaptive-alerting/issues/260

pom.xml
williewheeler: +1

PR #274, opened 2018-11-15
limited the no of results fetched by metric key containing query

karanshah03:


modelservice/src/main/java/com/expedia/adaptivealerting/modelservice/repo/MetricRepository.java
```diff
@@ -55,13 +55,14 @@
...
-    
-     /**
+
+    /**
      * Finds a list of metrics by its matching key.
      *
      * @param key Matching key.
      * @return List of metrics by its matching key
      */
+    @Query(nativeQuery = true, value = "SELECT * FROM metric m WHERE m.ukey LIKE %:key% LIMIT 20")
```
williewheeler: These should be ordered so it's easy to find what you're looking for.

PR #282, opened 2018-12-14
Created Params for HoltWinters anomaly detector including SeasonalityType.

mattcallanan:
First bite-sized commit preparing for new Holt-Winters Anomaly Detector.


anomdetect/src/main/java/com/expedia/adaptivealerting/anomdetect/holtwinters/SeasonalityTypeParser.java
```diff
@@ -0,0 +1,56 @@
...
+ * limitations under the License.
+ */
+package com.expedia.adaptivealerting.anomdetect.holtwinters;
+
+import java.util.Arrays;
+import java.util.List;
+import java.util.Optional;
+import java.util.stream.Collectors;
+
+public class SeasonalityTypeParser {
```
williewheeler: Java enums come with this already. See `Enum.valueOf()`.

mattcallanan: You're right, I didn't do enough whittling down of the original code. Thanks!

PR #284, opened 2018-12-18
Enhancement/model service new user

karanshah03:


modelservice/src/main/java/com/expedia/adaptivealerting/modelservice/security/ResourceServerConfig.java
```diff
@@ -39,9 +40,11 @@ public void configure(ResourceServerSecurityConfigurer resources) {
 
     @Override
     public void configure(HttpSecurity http) throws Exception {
-        http
-                .antMatcher("/**")
-                .authorizeRequests()
+
+        http.
+                anonymous().disable()
+                .requestMatchers().antMatchers("/api/**")
```
williewheeler: What happens if the user tries to access an internal (i.e., non-api) /** endpoint? Wouldn't we want those to be authenticated too?

I'm not totally following what's protected by what in this chain. Can you use indenting to communicate this more effectively? Look at how they do it here:

https://docs.spring.io/spring-security/site/docs/current/reference/html/jc.html#jc-httpsecurity

and here:

https://docs.spring.io/spring-security/site/docs/current/reference/html/jc.html#jc-form
https://docs.spring.io/spring-security/site/docs/current/reference/html/jc.html#jc-authorize-requests

modelservice/src/main/java/com/expedia/adaptivealerting/modelservice/security/SecurityConfig.java
```diff
@@ -46,22 +47,26 @@
...
-                .sessionManagement().sessionCreationPolicy(SessionCreationPolicy.STATELESS)
-                .and().authorizeRequests().antMatchers("/oauth/token")
-                .permitAll().anyRequest().authenticated();
+                .sessionManagement()
+                .sessionCreationPolicy(SessionCreationPolicy.STATELESS)
+                .and()
+                .authorizeRequests()
+                .antMatchers("/signUp").permitAll()
+                .antMatchers("/oauth/token").permitAll()
+                .anyRequest().authenticated();
```
williewheeler: See comment above about using indentation in the way that the Spring Security docs do.

modelservice/src/main/java/com/expedia/adaptivealerting/modelservice/security/ResourceServerConfig.java
karanshah03: Yeah. This is to secure the URL paths `/api/*` via an incoming OAuth2 token. We have a `configure` method in security class as well which contains the URL paths that should be secured in our application.

PR #294, opened 2019-01-10
Feature/add arima models anomaly detectors

dinilatgit:


williewheeler:
Dinil mentioned that he is still working on the test cases, so I am not reviewing this yet.

PR #301, opened 2019-01-22
Bug fix/constant threshold tails

karanshah03:


core/src/main/java/com/expedia/adaptivealerting/core/anomaly/AnomalyType.java
```diff
@@ -0,0 +1,17 @@
+package com.expedia.adaptivealerting.core.anomaly;
+
+import lombok.Data;
+
+@Data
+public class AnomalyType {
```
williewheeler: Why do we need a class and an enum here? Can we just have an enum?

Also I think we must already have an enum for CUSUM or one of the other detectors. Let's use the same left/right/two-tailed enum for all those detectors.

PR #348, opened 2019-02-18
Removed detector factory config from KafkaAnomalyDetectorManager

williewheeler:


anomdetect/src/main/java/com/expedia/adaptivealerting/anomdetect/DetectorLookup.java
```diff
@@ -45,6 +46,10 @@ public DetectorLookup() {
         detectorMap.put("pewma-detector", PewmaAnomalyDetector.class);
     }
     
+    public Set<String> getDetectorTypes() {
```
mattcallanan: I'd like to see a Unit Test for this class.
Can we use reflection to pick up the list of classes?

PR #357, opened 2019-02-19
Log full stack trace for detector classification errors

mattcallanan:


kafka/src/main/java/com/expedia/adaptivealerting/kafka/KafkaAnomalyDetectorManager.java
```diff
@@ -93,7 +96,13 @@ protected Topology buildTopology() {
...
-        // TODO Reduce this to the class, the message, and the first line of stack trace.
-        return e.toString();
+        return stackTraceString(e);
+    }
+
+    private String stackTraceString(Exception e) {
+        StringWriter stackTrace = new StringWriter();
+        PrintWriter pw = new PrintWriter(stackTrace);
+        e.printStackTrace(pw);
+        return stackTrace.toString();
```
williewheeler: Not sure it's required, but I think it's a good idea to close the writer. Maybe try-with-resources on the PrintWriter.

Issue #372, opened 2019-02-19
Write unit test for KafkaMultiClusterAnomalyToMetricMapper


williewheeler:
This class is missing a unit test. Please use `kafka-junit`.

williewheeler:
Done

Issue #367, opened 2019-02-19
Remove Spring from the Kafka Notifier app


williewheeler:
In general we would prefer not to use Spring for the Kafka Streams apps, as these are pretty lightweight and don't really warrant pulling Spring in. The Notifier app is currently based on Spring Boot. We would like to remove Spring from this particular app.

PR #374, opened 2019-02-20
Wrote KafkaMultiClusterAnomalyToMetricMapper unit test (#372)

williewheeler:
This involved creating some JSON-based serializers and deserializers,
along with their corresponding unit tests.

Also, I used kafka-junit to implement the unit a2m mapper unit test. After
doing so I noticed an interaction between the mock broker I am using and
the mock broker that the notifier app is using. I am not sure exactly what
the interaction is at this point, but I suspect it may be related to
the notifier tests setting system properties. Again that's just a guess.
At any rate I simply disabled the unit tests for the two notifier classes
as issue #367 is to remove Spring from the notifier. I will pick that
issue up after this one and see if that eliminates the interaction.

Finally please see my comments around JsonPojoSerde and the related
serializer and deserializer. I deprecated these. I explain in a comment in
JsonPojoDeserializer.


kafka/src/main/java/com/expedia/adaptivealerting/kafka/serde/JsonPojoDeserializer.java
```diff
@@ -24,6 +24,17 @@
...
+// configuration properties after it's already been typed. To me it looks like serializers and
+// deserializers should be concrete as they ProducerConfig and ConsumerConfig set them by
+// classname. For example
+//
+//   config.setProperty(ConsumerConfig.VALUE_DESERIALIZER_CLASS_CONFIG,
+//       "com.expedia.adaptivealerting.kafka.serde.MappedMetricDataJsonDeserializer");
+//
+// The config is just expecting a deserializer class that's already typed. [WLW]
+
+@Deprecated
```
djsutho: Would prefer this as a javadoc with `@deprecated` for the IDE support.

kafka/src/main/java/com/expedia/adaptivealerting/kafka/serde/JsonPojoSerde.java
```diff
@@ -21,6 +21,9 @@
 
 import java.util.Map;
 
+// Deprecated - see JsonPojoDeserializer
+
+@Deprecated
```
djsutho: see above

kafka/src/main/java/com/expedia/adaptivealerting/kafka/KafkaMultiClusterAnomalyToMetricMapper.java
```diff
@@ -17,124 +17,111 @@
...
 import org.apache.kafka.clients.producer.KafkaProducer;
 import org.apache.kafka.clients.producer.Producer;
-import org.apache.kafka.clients.producer.ProducerConfig;
 import org.apache.kafka.clients.producer.ProducerRecord;
+import org.apache.kafka.common.errors.WakeupException;
 
 import java.util.Collections;
-import java.util.Properties;
+
+// FIXME The input type is wrong. The anomalies topic contains MappedMetricData, not AnomalyResult. [WLW]
```
williewheeler: Oh, I fixed this. Lemme remove this comment.

kafka/src/main/java/com/expedia/adaptivealerting/kafka/serde/MappedMetricDataJsonDeserializer.java
```diff
@@ -0,0 +1,35 @@
...
+ *
+ * @author Willie Wheeler
+ */
+public class MappedMetricDataJsonDeserializer extends AbstractJsonDeserializer<MappedMetricData> {
+    
+    public MappedMetricDataJsonDeserializer() {
+        super(MappedMetricData.class);
+        getObjectMapper()
+                .registerModule(new MetricsJavaModule())
+                .configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
```
djsutho: Is it worth making
```
 .configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
.configure(SerializationFeature.FAIL_ON_EMPTY_BEANS, false);
```
the default in the super class

kafka/src/main/java/com/expedia/adaptivealerting/kafka/KafkaMultiClusterAnomalyToMetricMapper.java
```diff
@@ -17,124 +17,109 @@
...
-                        metricProducer.send(toMetricDataRecord(record));
-                    }
-                });
-            } catch (Exception e) {
-                log.error("Exception while processing anomalies", e);
+        try {
+            while (true) {
+                val records = anomalyConsumer.poll(POLL_PERIOD);
+                log.trace("numRecords={}", records.count());
+                records.forEach(record -> metricProducer.send(toMetricDataRecord(record)));
```
djsutho: Are we no-longer worried about breaking from the while loop? Do we just wait for app to restart? What if we have a bad record on the topic

PR #379, opened 2019-02-21
added support to preview anomalies for a given detector and metric

karanshah03:


modelservice/src/main/java/com/expedia/adaptivealerting/modelservice/providers/graphite/GraphiteRequest.java
```diff
@@ -0,0 +1,36 @@
...
+@NoArgsConstructor
+@AllArgsConstructor
+public class GraphiteRequest {
+    private String target;
+
+    public Map<String, Object> toParams() {
+        Map<String, Object> params = new HashMap<>();
+        params.put("target", target);
+        return params;
+    }
```
williewheeler: Not necessary to generate a param map whose JSON is identical to the `GraphiteRequest` itself. Just use the `GraphiteRequest`. Could just pass a singleton map as well.

modelservice/src/main/java/com/expedia/adaptivealerting/modelservice/providers/graphite/GraphiteMetricSource.java
```diff
@@ -0,0 +1,57 @@
...
+
+    private RestTemplate restTemplate = new RestTemplate();
+
+    @Override
+    public List<MetricSourceResult> getMetricData(String metricName) {
+        GraphiteProperties props = BeanUtil.getBean(GraphiteProperties.class);
+        GraphiteRequest request = new GraphiteRequest(metricName);
+        Map<String, Object> params = request.toParams();
+
+        GraphiteResult graphiteResult = restTemplate.getForObject(props.getUrlTemplate(), GraphiteResult[].class, params)[0];
```
williewheeler: Could just pass in

`Collections.singletonMap("target", metricName")`

If you want to keep the `GraphiteRequest`, that's OK too -- but just pass the request itself in, and get rid of the `toParams()` method. Not needed.

modelservice/src/main/java/com/expedia/adaptivealerting/modelservice/service/AnomalyRequest.java
```diff
@@ -0,0 +1,27 @@
...
+ */
+package com.expedia.adaptivealerting.modelservice.service;
+
+import lombok.Data;
+
+import java.util.Map;
+
+@Data
+public class AnomalyRequest {
+    private String hash;
```
williewheeler: This should be `metricId`, not `hash`. The fact that we generate the ID as a hash is an implementation detail. And we'd want it to be `metricId` so that it doesn't sound like anomaly request ID.

But we're getting ready to remove metrics from model service, so we should avoid building metric IDs into the system, that's what this is supposed to represent.

karanshah03: Do you mean the metricId which we store in model service DB?

williewheeler: See https://graphite.readthedocs.io/en/latest/tags.html

Can we try this?

modelservice/src/main/java/com/expedia/adaptivealerting/modelservice/util/BeanUtil.java
```diff
@@ -0,0 +1,41 @@
...
+import org.springframework.beans.BeansException;
+import org.springframework.context.ApplicationContext;
+import org.springframework.context.ApplicationContextAware;
+import org.springframework.stereotype.Service;
+
+/**
+ * <p>
+ * We cannot inject any Spring-managed bean in any of the provider classes because these are instantiated by JPA before Spring injects anything into it.
+ * This is a utility class to fetch any bean according to our requirements.
+ * </p>
```
williewheeler: I'm not sure I'm following. The SPI is a metric source SPI and the current provider is a Graphite provider. I would not expect JPA to load anything related to metrics sources at all.

modelservice/src/test/java/com/expedia/adaptivealerting/modelservice/test/ObjectMother.java
```diff
@@ -0,0 +1,64 @@
...
+    public static ObjectMother instance() {
+        return MOM;
+    }
+
+    private ObjectMother() {
+    }
+
+    public GraphiteResult[] getGraphiteData() {
+        GraphiteResult[] results = new GraphiteResult[1];
+        GraphiteResult result = new GraphiteResult();
```
williewheeler: Stylistically, try to keep a declaration close to the code that actually uses it. For example here is would be clearer to build the datapoints first, then declare and build the result, and finally declare and build the results.

Like I say this is stylistic guidance, so it's your call.

modelservice/src/test/java/com/expedia/adaptivealerting/modelservice/test/ObjectMother.java
```diff
@@ -0,0 +1,64 @@
...
+        String[][] datapoints = new String[2][2];
+        datapoints[0][0] = "78.0";
+        datapoints[0][1] = "1548829800";
+        datapoints[1][0] = "81.0";
+        datapoints[1][1] = "1548830400";
+        result.setDatapoints(datapoints);
+        result.setTags(new Tags());
+        result.setTarget("karmalab.stats.gauges.AirBoss.chelappabo004_karmalab_net.java.nio.BufferPool.mapped.TotalCapacity30");
```
williewheeler: Please keep Expedia-specific data out of the OSS repo.

modelservice/src/test/java/com/expedia/adaptivealerting/modelservice/test/ObjectMother.java
```diff
@@ -0,0 +1,64 @@
...
+        request.setDetectorParams(detectorParams);
+        request.setDetectorType("constant-detector");
+        request.setHash("1.3dec7f4218c57c1839147f8ca190ed55");
+        return request;
+    }
+
+    public Metric getMetric() {
+        Metric metric = new Metric();
+        metric.setHash("1.3dec7f4218c57c1839147f8ca190ed55");
+        metric.setKey("karmalab.stats.gauges.AirBoss.chelappabo004_karmalab_net.java.nio.BufferPool.mapped.TotalCapacity30");
```
williewheeler: Please keep Expedia-specific data out of the OSS repo.

modelservice/src/test/java/com/expedia/adaptivealerting/modelservice/test/ObjectMother.java
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    public GraphiteResult[] getGraphiteData() {
+        GraphiteResult[] results = new GraphiteResult[1];
+        GraphiteResult result = new GraphiteResult();
+        String[][] datapoints = new String[2][2];
+        datapoints[0][0] = "78.0";
+        datapoints[0][1] = "1548829800";
+        datapoints[1][0] = "81.0";
+        datapoints[1][1] = "1548830400";
```
williewheeler: Given that we defined our own `GraphiteResult` class here, I would not expect the data values and epoch timestamps to be strings. I would expect numerical values (say double/long). If Graphite expects string then map the numbers to strings at serialization time.

modelservice/src/test/java/com/expedia/adaptivealerting/modelservice/test/ObjectMother.java
```diff
@@ -0,0 +1,64 @@
...
+
+    public MetricSourceResult getMetricData() {
+        MetricSourceResult result = new MetricSourceResult();
+        result.setDataPoint(78.0);
+        result.setEpochSecond(1548830400);
+        return result;
+    }
+
+    public AnomalyRequest getAnomalyRequest() {
+        AnomalyRequest request = new AnomalyRequest();
```
williewheeler: Like above consider moving this closer to the request code below.

karanshah03:
Closing this PR. Raising a new one for this change.

PR #381, opened 2019-02-23
Stop pushing non-anomalous metrics to the anomalies topic.

williewheeler:
The point of the anomalies topic is to notify downstreams of anomalous
metrics, so it doesn't make sense to send non-anomalous metrics.

I've left logging non-anomalous metrics in place for now. It is unclear
whether we want to keep that (due to the volume), but that's a follow-up
discussion with the team.

In the KafkaAnomalyDetectorMapper, I replaced the JsonPojoSerde (output
value serde) with a MappedMetricDataJsonSerde, as the former is
deprecated. Behaviorally, the MMDJsonSerde returns null when
deserialization fails, instead of throwing an exception. This is in line
with the guidance here:

https://stackoverflow.com/questions/51136942/how-to-handle-serializationexception-after-deserialization

Finally, I added a few unit test cases and refactored the manager and
mapper a bit, mostly to clarify the stream processing logic in those two
components.

### DIFF
--- a/kafka/src/main/java/com/expedia/adaptivealerting/kafka/KafkaAnomalyDetectorManager.java
+++ b/kafka/src/main/java/com/expedia/adaptivealerting/kafka/KafkaAnomalyDetectorManager.java
@@ -16,6 +16,7 @@
 package com.expedia.adaptivealerting.kafka;
 
 import com.expedia.adaptivealerting.anomdetect.DetectorManager;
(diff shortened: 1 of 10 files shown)


williewheeler:
Karan mentioned that if we don't push normals, it may be hard to let people know when a metric has "recovered" (gone from anomalous to not). Closing for now.

PR #382, opened 2019-02-24
Cleanup/detector manager

williewheeler:


kafka/src/test/java/com.expedia.adaptivealerting.kafka/KafkaDetectorManagerTest.java
```diff
@@ -48,6 +48,7 @@
     private static final String KAFKA_KEY = "some-kafka-key";
     private static final String INBOUND_TOPIC = "mapped-metrics";
     private static final String OUTBOUND_TOPIC = "anomalies";
+    private static final String INVALID_VALUE = "invalid-value";
```
djsutho: Should this be `INVALID_METRIC_VALUE` not 100% sure about its usage.

kafka/src/test/java/com.expedia.adaptivealerting.kafka/KafkaDetectorManagerTest.java
```diff
@@ -110,28 +143,59 @@ private void initConfig() {
...
+    
+    private void publishesAnomaly(MappedMetricData metric, AnomalyLevel anomalyLevel) {
+        val anomalyRecord = getAnomalyRecord(metric);
+        val anomaly = anomalyRecord.value();
+        assertNotNull(anomaly);
+        assertEquals(anomalyLevel, anomaly.getAnomalyResult().getAnomalyLevel());
+    }
+    
+    @SuppressWarnings("unused")
+    private void doesNotPublishAnomaly(MappedMetricData metric) {
```
djsutho: Is this for future use or am I missing its purpose?

kafka/src/test/java/com.expedia.adaptivealerting.kafka/KafkaDetectorMapperTest.java
```diff
@@ -98,25 +101,27 @@ public void testMetricDataToMappedMetricData() {
...
+     * See also https://stackoverflow.com/questions/51136942/how-to-handle-serializationexception-after-deserialization
      */
-    @Test(expected = StreamsException.class)
-    public void testFailsOnDeserializationException() {
-        logAndFailDriver.pipeInput(stringFactory.create(INBOUND_TOPIC, KAFKA_KEY, "invalid_input"));
-        logAndFailDriver.readOutput(OUTBOUND_TOPIC, stringDeserializer, mmdDeserializer);
+    @Test
+    public void nullOnDeserExceptionWithLogAndFailDriver() {
+        logAndFailDriver.pipeInput(stringFactory.create(INBOUND_TOPIC, KAFKA_KEY, INVALID_INPUT));
+        val record = logAndFailDriver.readOutput(OUTBOUND_TOPIC, stringDeserializer, mmdDeserializer);
```
djsutho: Looks like we are duplicating some tests from `KafkaDetectorManagerTest` can we either test the super class separately or inherit from a common test class?

kafka/src/test/java/com.expedia.adaptivealerting.kafka/KafkaDetectorManagerTest.java
williewheeler: Well, previously I was verifying that we _don't_ publish certain conditions, but after we changed it, I thought I'd keep this logic around in case we wanted to suppress something. Let me know if you prefer that I simply remove it. I can go either way.

kafka/src/test/java/com.expedia.adaptivealerting.kafka/KafkaDetectorMapperTest.java
williewheeler: Sure, I'll look at that. Thanks.

kafka/src/test/java/com.expedia.adaptivealerting.kafka/KafkaDetectorManagerTest.java
williewheeler: Will clarify it, thanks.

williewheeler: Updated.

kafka/src/test/java/com.expedia.adaptivealerting.kafka/KafkaDetectorMapperTest.java
williewheeler: I didn't end up doing this. The test methods in question have several params (topics, deserializers, inputs, drivers) and by the time that all goes into an abstract test class (or even utility), the resulting code seems more complicated than what's currently here. I did streamline the implementations a bit to use utility methods for the individual test classes. My thought is that if we add those test cases to other tests then it might tip the balance in favor of extracting those. Let me know if you have any concerns. If you want me to extract them I will.

PR #389, opened 2019-02-28
a2m bugfix

williewheeler:


kafka/src/main/java/com/expedia/adaptivealerting/kafka/KafkaMultiClusterAnomalyToMetricMapper.java
```diff
@@ -100,9 +100,14 @@ public void run() {
         try {
             while (true) {
                 try {
-                    val records = anomalyConsumer.poll(POLL_PERIOD);
-                    log.trace("Read {} records from topic={}", records.count(), anomalyTopic);
-                    records.forEach(record -> metricProducer.send(toMetricDataRecord(record)));
+                    val anomalyRecords = anomalyConsumer.poll(POLL_PERIOD);
+                    log.trace("Read {} anomalyRecords from topic={}", anomalyRecords.count(), anomalyTopic);
+                    anomalyRecords.forEach(record -> {
```
bibinss: probably we can add a filter here to pass in only valid records for transformation instead of returning null from AnomalyToMetricTransformer->transform()

williewheeler: +1, will update it. Thanks for the catch.

PR #391, opened 2019-03-02
mc-a2m-mapper must handle tags with null values

williewheeler:
In production we see metrics with null tag values, for whatever reason.
Metricstank and Metrics 2.0 don't support this, so currently such metrics
generate IllegalArgumentExceptions and the mapper gets stuck on them.

This commit catches the IllegalArgumentException and responds by returning
null for offending metrics.


anomdetect/src/test/java/com/expedia/adaptivealerting/anomdetect/AnomalyToMetricMapperTest.java
```diff
@@ -28,6 +29,7 @@
 import static com.expedia.adaptivealerting.anomdetect.AnomalyToMetricMapper.AA_DETECTOR_UUID;
 import static org.junit.Assert.assertTrue;
 
+@Slf4j
```
bibinss: is this required?

williewheeler: Yeah, I'm displaying an error message if there's an exception. I added this after your question.

PR #406, opened 2019-03-09
Create CODE_OF_CONDUCT.md

williewheeler:


CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1,76 @@
...
+obligated to maintain confidentiality with regard to the reporter of an incident.
+Further details of specific enforcement policies may be posted separately.
+
+Project maintainers who do not follow or enforce the Code of Conduct in good
+faith may face temporary or permanent repercussions as determined by other
+members of the project's leadership.
+
+## Attribution
+
+This Code of Conduct is adapted from the [Contributor Covenant][homepage], version 1.4,
```
adriancole: nice to attribute

CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1,76 @@
...
+Examples of behavior that contributes to creating a positive environment
+include:
+
+* Using welcoming and inclusive language
+* Being respectful of differing viewpoints and experiences
+* Gracefully accepting constructive criticism
+* Focusing on what is best for the community
+* Showing empathy towards other community members
+
+Examples of unacceptable behavior by participants include:
```
adriancole: I always wondered why "persistent misinformation" has never gotten onto code of conducts. For example, misattribution even once corrected, self-glorification, "shameless plugs" clearly aimed to not help the community but rather drift it to somewhere etc.

williewheeler: Hey @adriancole . Feel free to submit anything you think would make sense, both on the positive and negative behaviors.

williewheeler: In fact, the whole conversation around adding a code of conduct followed your tweet (along with some internal discussions) around using @ author tags in the code. We decided to remove those just because they smelled too much of file ownership, where we want to have a community-ownership model. We haven't worked that into the code of conduct yet but I see some alignment with what you say about putting the community first. So if you have some suggestions on how best to incorporate that thinking into the code please share those.

adriancole: sadly I've still seen no example I can find great about this, and haven't put enough brain to figure it out either.

CoC is a tough one in oss as you literally say below corrective action will be taken! who will do that, right? who will police etc? This is one reason I hesitate.. it might be better to say corrective action "may" take place unless a project has the resources to make promises. Given that generally common shaky foundation of CoC, it is hard to layer on other rules as sensible as I might think they can be because, often the commercial offenders are in power positions that would make it easy for them to dodge.. there's no independent watchdog party that can help smaller parties deal with offenders. If anything routine offenders are given more money.

In other words, when I start thinking about the practicalities of CoC.. unless they are corporate backed, which might be the case here, it feels empty promises especially on the moderation and failing that enforcement side.

It has been my experience that volunteers rarely want to walk into CoC related activity when choosing between that and watching TV

adriancole: PS on author files, I think that can be more directly related in a CONTRIBUTING file, like that here is the codestyle which does not permit author tags. I think CoC are the darker side of things, CONTRIBUTING can be an easy place to take care of rules that are in some ways arbitrary and arguable, but not in themselves an argument

williewheeler: Re: enforcement of conduct--I don't take this to commit me or anybody else to an ongoing monitoring/policing responsibility. We all have better things to do. Rather it's just letting people know what's considered good vs. poor behavior, and that the maintainers make that decision and can delete stuff. This CoC is explicitly saying that it's _not_ going to be an appear to an independent party. I don't see any reason for undue skepticism here. If somebody posts sexist content, racial slurs, personal insults--I have no problem deleting that stuff.

williewheeler: I suspect you have some other less cut-and-dried cases in mind--like erased attribution--and it's not clear what "enforcement" looks like in that situation. If I have you right, I agree.

adriancole:
FWIW I am happy you've thought this through. Hopefully the only extent of
impact is just whether or not to delete something or freeze a thread.
considering this is a repo scoped CoC anyway probably that makes sense.

go for it!

PR #411, opened 2019-03-11
ad-manager to use only detetctor uuid for detector lookup

bibinss:


anomdetect/src/main/java/com/expedia/adaptivealerting/anomdetect/DetectorManager.java
```diff
@@ -88,12 +88,13 @@ private AnomalyDetector detectorFor(MappedMetricData mappedMetricData) {
...
         val detectorUuid = mappedMetricData.getDetectorUuid();
-        val detectorType = mappedMetricData.getDetectorType();
         val metricDef = mappedMetricData.getMetricData().getMetricDefinition();
         
         AnomalyDetector detector = cachedDetectors.get(detectorUuid);
         if (detector == null) {
-            val detectorMeta = new DetectorMeta(detectorUuid, detectorType);
+            //FIXME detector is looked up using detectorUuid. We probably need to get rid of DetectorMeta class once we
+            //have tag based detector lookup logic implemented with in AA.
+            val detectorMeta = new DetectorMeta(detectorUuid, null);
```
williewheeler: If you are nulling out the type, wouldn't it make more sense to simply remove the type from the meta?

And if the meta has only one field, can we just get rid of the meta altogether?

bibinss: yes. i can remove the meta type fully. will update the PR

PR #434, opened 2019-03-16
Enhancement/exception improvements

williewheeler:


anomdetect/src/main/java/com/expedia/adaptivealerting/anomdetect/source/DetectorSource.java
```diff
@@ -43,15 +43,19 @@
...
      * @param metricDef The metric.
      * @return The detector UUIDs.
+     * @throws DetectorException if there's a problem finding the detectors
      */
     List<UUID> findDetectorUuids(MetricDefinition metricDef);
 
     /**
-     * Finds the detector for a given detector and, optionally, metric.
+     * Finds the detector for a given detector and, optionally, metric. We expect the detector to exist and to be
+     * well-defined, so we throw a {@link DetectorException} if that's not the case.
```
mattcallanan: Just a nit, should be:

```
... so we throw a {@link DetectorNotFoundException} or {@link DetectorException} if that's not the case.
```

or don't mention the exceptions at all given they're documented below.

williewheeler: I'll update

PR #441, opened 2019-03-18
scheduled model-service polling to get modified detectors and deletin…

shsethi:
…g them from cachedDetectors


williewheeler:
Please add unit tests

anomdetect/src/main/java/com/expedia/adaptivealerting/anomdetect/DetectorManager.java
```diff
@@ -37,13 +38,33 @@
...
+
+    private void initScheduler() {
+        scheduler.scheduleWithFixedDelay(() -> {
+            try {
+                this.detectorMapRefresh();
+            } catch (Exception e) {
+                log.error("Error refreshing detectors", e);
+            }
+        }, 1, detectorRefreshTimePeriod, TimeUnit.MINUTES);
+    }
```
williewheeler: If we get an exception, do we just lose the refresh until we restart the manager? What kind of exception are we expecting here, and is it something that we can recover from without losing refreshes entirely?

anomdetect/src/main/java/com/expedia/adaptivealerting/anomdetect/DetectorManager.java
```diff
@@ -96,4 +117,21 @@ private AnomalyDetector detectorFor(MappedMetricData mappedMetricData) {
...
+     */
+    List<UUID> detectorMapRefresh() {
+
+        List<UUID> updatedDetectors = new ArrayList<>();
+        detectorSource.findUpdatedDetectors(detectorRefreshTimePeriod).forEach(key -> {
+            updatedDetectors.add(key);
+            cachedDetectors.remove(key);
+        });
+
+        log.info("Removed detectors on refresh : {}"+updatedDetectors);
```
williewheeler: The `+` should be a `,`.

anomdetect/src/main/java/com/expedia/adaptivealerting/anomdetect/util/DetectorResource.java
```diff
@@ -30,4 +30,5 @@
 public class DetectorResource {
     private String uuid;
     private ModelTypeResource type;
+    private Boolean enabled;
```
williewheeler: Can you clarify what this is for, and how it gets set?

anomdetect/src/main/java/com/expedia/adaptivealerting/anomdetect/util/ModelServiceConnector.java
```diff
@@ -146,4 +150,29 @@ public ModelResource findLatestModel(UUID detectorUuid) {
 
         return modelResourceList.get(0);
     }
+
+    public DetectorResources findUpdatedDetectors(int timePeriod) {
+        isTrue(timePeriod > 0, "timePeriod can't be negative");
```
williewheeler: Consider "timePeriod must be strictly positive". The current error message allows 0, but in fact that's not allowed.

anomdetect/src/main/java/com/expedia/adaptivealerting/anomdetect/util/DetectorResource.java
shsethi: Detectors are get disabled through and endpoint exposed by model-service, default is true, it's set to false when someones deletes detector from UI

anomdetect/src/main/java/com/expedia/adaptivealerting/anomdetect/DetectorManager.java
shsethi: This is to catch DetectorException's and any other Null pointer exceptions that might occur downstream. We don't loose the refresh when exception is caught.

PR #452, opened 2019-03-24
Removed unused classes + fields in preparation for detector enhancement.

williewheeler:
Removed:
- TempHaystackAwareDetectorSource and test
- Unused ModelResource fields (id, uuid)

Changed:
- Replaced Timestamp in Model + ModelResource with Date since there's no
  need for the SQL-specific version here, and we don't want to expose SQL
  types in a REST resource anyway.


anomdetect/src/test/java/com/expedia/adaptivealerting/anomdetect/DetectorManagerTest.java
```diff
@@ -145,14 +151,8 @@ private void initTestObjects() {
     private void initDependencies() {
         when(detector.classify(goodMetricData)).thenReturn(anomalyResult);
 
-        when(detectorSource.findDetectorTypes())
-                .thenReturn(Collections.singleton(DETECTOR_TYPE));
-
-        when(detectorSource.findDetector(any(UUID.class), eq(goodDefinition)))
-                .thenReturn(detector);
-        when(detectorSource.findDetector(any(UUID.class), eq(badDefinition)))
```
mattcallanan: Any reason not to test this?  This has caused coverage of DetectorManager coverage to drop by 8%:
https://codecov.io/gh/ExpediaDotCom/adaptive-alerting/pull/452/src/anomdetect/src/main/java/com/expedia/adaptivealerting/anomdetect/DetectorManager.java#L120

williewheeler: Thanks, this was a miss.

PR #454, opened 2019-03-26
Enhancement/interval forecasts

williewheeler:


anomdetect/src/main/java/com/expedia/adaptivealerting/anomdetect/forecast/interval/IntervalForecast.java
```diff
@@ -0,0 +1,52 @@
...
+    private double upperWeak;
+    private double lowerWeak;
+    private double lowerStrong;
+
+    @JsonCreator
+    public IntervalForecast(
+            @JsonProperty("upperStrong") double upperStrong,
+            @JsonProperty("upperWeak") double upperWeak,
+            @JsonProperty("lowerWeak") double lowerWeak,
+            @JsonProperty("lowerStrong") double lowerStrong) {
```
djsutho: Do we need the `@JsonProperty` when the name is the same as the parameter?
Are these also always required?

williewheeler: I'm not sure actually. In Spring you do not as long as debug info isn't disabled. In Jackson I don't know. I'll look it up.

PR #466, opened 2019-04-07
Refactored EwmaDetector into ForecastingDetector

williewheeler:
This commit moves us toward being able to mix and match point forecasting
algos with interval forecasting algos. Here we replace the EwmaDetector
(which coupled the EWMA point forecast algo with the exponential version
of the Welford stdev algo) with a ForecastingDetector, which takes a point
forecasting algo, an interval forecasting algo and a tail type.


anomdetect/src/test/java/com/expedia/adaptivealerting/anomdetect/forecast/point/PewmaPointForecasterTest.java
```diff
@@ -66,70 +60,57 @@ public void pewmaCloseToEwmaWithZeroBeta() throws IOException {
...
-        val ewmaDetector = new EwmaDetector();
-        ewmaDetector.init(UUID.randomUUID(), ewmaParams, AnomalyType.TWO_TAILED);
+        val ewmaPointForecaster = new EwmaPointForecaster(ewmaParams.toPointForecasterParams());
 
         int rowCount = 1;
         while (testRows.hasNext()) {
             val observed = Float.parseFloat(testRows.next()[0]);
-
-            val ewmaStdDev = sqrt(ewmaDetector.getVariance());
+//            val ewmaStdDev = sqrt(ewmaDetector.getVariance());
```
mattcallanan: Remove or reinstate comments in this method

anomdetect/src/test/java/com/expedia/adaptivealerting/anomdetect/forecast/ForecastingDetectorTest.java
```diff
@@ -0,0 +1,77 @@
...
+        this.detectorUnderTest = new ForecastingDetector(
+                UUID.randomUUID(),
+                pointForecaster,
+                intervalForecaster,
+                AnomalyType.TWO_TAILED);
+    }
+
+    @Test
+    public void testClassify() {
+        val metricDef = new MetricDefinition("my-metric");
```
mattcallanan: Suggest pulling "my-metric" out as a constant e.g. `DUMMY_METRIC_KEY = "dummy-metric"`.  
The "dummy" label can help demonstrate there's no magic going on behind the scenes and the test doesn't rely on the value.

anomdetect/src/test/java/com/expedia/adaptivealerting/anomdetect/forecast/ForecastingDetectorTest.java
```diff
@@ -0,0 +1,77 @@
...
+        val result = detectorUnderTest.classify(metricData);
+        assertNotNull(result);
+    }
+
+    @Test(expected = IllegalArgumentException.class)
+    public void testClassify_nullMetricData() {
+        detectorUnderTest.classify(null);
+    }
+
+    private void initDependencies() {
```
mattcallanan: Consider `setupExpectations` as name for this method.

anomdetect/src/test/java/com/expedia/adaptivealerting/anomdetect/forecast/ForecastingDetectorTest.java
```diff
@@ -0,0 +1,77 @@
...
+                UUID.randomUUID(),
+                pointForecaster,
+                intervalForecaster,
+                AnomalyType.TWO_TAILED);
+    }
+
+    @Test
+    public void testClassify() {
+        val metricDef = new MetricDefinition("my-metric");
+        val metricData = new MetricData(metricDef, 100.0, Instant.now().getEpochSecond());
```
mattcallanan: Also, consider naming `dummyMetricDef` and `dummyMetricData`.
And pulling out all magic number parameters as `DUMMY_...` constants.

mattcallanan: Consider pulling first two lines out as a `dummyMetricData()` method

anomdetect/src/main/java/com/expedia/adaptivealerting/anomdetect/forecast/point/PewmaPointForecaster.java
```diff
@@ -129,17 +106,36 @@ private void updateEstimates(double value) {
...
         }
         return (1.0 - params.getBeta() * pt) * this.adjAlpha;
     }
+
+    @Data
+    @Accessors(chain = true)
+    public static class Params {
+
+        /**
+         * Smoothing param.
```
mattcallanan: Let's add a TODO for "Describe why we chose 0.15, 1.0, 0.0, 30 as defaults for this class."

anomdetect/src/main/java/com/expedia/adaptivealerting/anomdetect/forecast/ForecastingDetector.java
```diff
@@ -0,0 +1,95 @@
...
+import lombok.NonNull;
+import lombok.RequiredArgsConstructor;
+import lombok.val;
+
+import java.util.UUID;
+
+import static com.expedia.adaptivealerting.core.util.AssertUtil.notNull;
+
+@RequiredArgsConstructor
+public class ForecastingDetector implements Detector {
```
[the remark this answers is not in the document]
williewheeler: Added Javadoc comments. Didn't add any UML for now since the code is still in flux, but no objection to adding that later.

anomdetect/src/test/java/com/expedia/adaptivealerting/anomdetect/forecast/point/PewmaPointForecasterTest.java
williewheeler: Done

anomdetect/src/test/java/com/expedia/adaptivealerting/anomdetect/forecast/ForecastingDetectorTest.java
williewheeler: I redid this to pull the metric def from an object mother.

anomdetect/src/test/java/com/expedia/adaptivealerting/anomdetect/forecast/ForecastingDetectorTest.java
williewheeler: If we do this then we need to do it throughout as all my test classes use `initDependencies` and `initTestObjects`. In general I like to separate those two because the fixture code can get pretty large and tough to follow. Happy to discuss further though. I like using "setup" as opposed to "init" here since it's a setup() delegation though.

anomdetect/src/test/java/com/expedia/adaptivealerting/anomdetect/forecast/ForecastingDetectorTest.java
williewheeler: I guess that doesn't really address the issue though. Maybe I can rename the variable to `dummyMetricDef`.

williewheeler: Hm. The variables are test-level fields in a lot of cases. Not sure we want to have to name all these "dummy-this", "dummy-that". My suggestion would be that we pick an example and then show me what you have in mind, and let's see whether it's a general pattern we want to adopt and how we want to do so.

FWIW I do like the idea of making it clearer when something is a dummy object. So not necessarily against dummy-this etc.

anomdetect/src/main/java/com/expedia/adaptivealerting/anomdetect/forecast/point/PewmaPointForecaster.java
williewheeler: We probably want to be a little selective here. If the paper recommends certain defaults then I think that's worth documenting. In the case of the initial mean estimate, it was just that 0.0 is a less-strange default than 4.3 would have been. I'll add the TODO though.

PR #471, opened 2019-04-09
Refactor/polish

williewheeler:


anomdetect/src/test/java/com/expedia/adaptivealerting/anomdetect/comp/legacy/HoltWintersDetectorTest.java
```diff
@@ -42,6 +42,7 @@
 /**
  * Tests Holt-Winters functionality by comparing with data generated from Hyndman's R "fpp2" library - see GenerateAustouristsTests.R
  */
+@Deprecated
```
djsutho: Is the test actually deprecated or just the code it's testing? Reason I care is I like to see explanations going with `@Deprecated` so user knows what they should be doing instead.

anomdetect/src/main/java/com/expedia/adaptivealerting/anomdetect/comp/legacy/LegacyDetectorFactory.java
```diff
@@ -46,30 +41,36 @@
 @Slf4j
 @Deprecated
```
djsutho: Would be nice to link to whatever is replacing this class or some sort of description.

anomdetect/src/test/java/com/expedia/adaptivealerting/anomdetect/comp/legacy/HoltWintersDetectorTest.java
williewheeler: Well, both. When the code being tested goes away I would remove the test too. But maybe "deprecated" is odd to apply to unit tests since they aren't APIs, and the deprecated label is for letting people know not to use something. Is that what you have in mind?

djsutho: Pretty much. I expect when I see an `@Deprecated` some sort of info on the alternative or why (otherwise it's documentation without usable documentation). I personally wouldn't call a test deprecated (I'm not trying to warn someone not to use the test) but others may feel differently.

anomdetect/src/main/java/com/expedia/adaptivealerting/anomdetect/comp/legacy/LegacyDetectorFactory.java
williewheeler: It doesn't exist yet, but I will.

anomdetect/src/test/java/com/expedia/adaptivealerting/anomdetect/comp/legacy/HoltWintersDetectorTest.java
williewheeler: OK. That makes sense to me. Thanks.

PR #472, opened 2019-04-09
exposed an endpoint in model servie to preview anomalies

karanshah03:


modelservice/src/main/java/com/expedia/adaptivealerting/modelservice/providers/graphite/Tags.java
```diff
@@ -0,0 +1,24 @@
...
+ */
+package com.expedia.adaptivealerting.modelservice.providers.graphite;
+
+import lombok.Data;
+
+@Data
+public class Tags {
+
+    private String name;
+}
```
williewheeler: `Tags` should not be a class with a hardcoded tag name, since the actual tag names are arbitrary. Here is a real example:

```
"tags": {
  "lob": "flights",
  "what": "bookings",
  "m_account": [redacted],
  "name": "booking",
  "region": "expedia-all",
  "m_region": [redacted],
  "m_application": [redacted]
}
```

Instead `GraphiteResult` should contain a map called `tags` and allow any dynamic (i.e. not hardcoded) set of tags.

modelservice/src/main/java/com/expedia/adaptivealerting/modelservice/providers/graphite/GraphiteResult.java
```diff
@@ -0,0 +1,32 @@
...
+import lombok.ToString;
+
+@Data
+@NoArgsConstructor
+@AllArgsConstructor
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class GraphiteResult {
+    private String[][] datapoints;
+    private String target;
+    private Tags tags;
```
williewheeler: This should be a `Map<String, Object>` of tags. See below.

modelservice/src/main/java/com/expedia/adaptivealerting/modelservice/providers/graphite/GraphiteMetricSource.java
```diff
@@ -0,0 +1,63 @@
...
+import java.util.List;
+import java.util.Map;
+
+
+/**
+ * Graphite metric source
+ */
+@Slf4j
+@Service
+@Configurable
```
williewheeler: I do not believe this to be a proper use of `@Configurable`. `@Configurable` is for doing dependency injection into domain objects, and generally involves AOP and weaving. Since this is a service, we should use standard autowiring here.

It may be that you're doing something tricky here to make this a pluggable provider. If that's the case, it would help me to better understand what you are doing. We can chat about it.

modelservice/src/main/java/com/expedia/adaptivealerting/modelservice/util/BeanUtil.java
```diff
@@ -0,0 +1,41 @@
...
+import org.springframework.stereotype.Service;
+
+/**
+ * <p>
+ * We cannot inject any Spring-managed bean in any of the provider classes because these are instantiated by JPA before Spring injects anything into it.
+ * This is a utility class to fetch any bean according to our requirements.
+ * </p>
+ */
+@Service
+public class BeanUtil implements ApplicationContextAware {
```
williewheeler: Not excited about this class but I understand you're doing it as a hack. I'll take a closer look to see where you are using this and to see whether there are alternatives.

This should be a `@Component` rather than a `@Service`, though. A service is more for domain logic, and that's not what this is doing.

modelservice/src/main/java/com/expedia/adaptivealerting/modelservice/web/AnomalyController.java
```diff
@@ -0,0 +1,38 @@
...
+
+import java.util.List;
+
+@RestController
+public class AnomalyController {
+
+    @Autowired
+    private AnomalyService anomalyService;
+
+    @PostMapping(path = "/api/anomalies", consumes = "application/json", produces = "application/json")
```
williewheeler: If we use the `/api` prefix, let's attach that to the controller itself so we don't have to repeat it for each handler method.

modelservice/src/test/java/com/expedia/adaptivealerting/modelservice/providers/graphite/GraphiteMetricSourceTest.java
```diff
@@ -0,0 +1,74 @@
...
+        assertNotNull(metricSourceResults);
+        assertEquals(2, metricSourceResults.size());
+    }
+
+    private void initTestObjects() {
+        ObjectMother mom = ObjectMother.instance();
+        this.beanUtil = new BeanUtil();
+        beanUtil.setApplicationContext(ctx);
+        results.add(mom.getMetricData());
+    }
```
williewheeler: Small stylistic suggestion. The following order would make more sense:

```
        this.beanUtil = new BeanUtil();
        beanUtil.setApplicationContext(ctx);

        ObjectMother mom = ObjectMother.instance();
        results.add(mom.getMetricData());
```

The two outer lines are related to each other, so there shouldn't be other code sitting in between them.

modelservice/src/test/java/com/expedia/adaptivealerting/modelservice/providers/graphite/GraphiteMetricSourceTest.java
```diff
@@ -0,0 +1,74 @@
...
+
+    private void initDependencies() {
+        ObjectMother mom = ObjectMother.instance();
+        GraphiteProperties props = new GraphiteProperties();
+        props.setUrlTemplate("https://graphiteURl/render?from=-24hours&format=json&maxDataPoints=144&target={target}");
```
williewheeler: Use consistent case: either `graphiteURL` or `graphiteUrl`, but not `graphiteURl`.

modelservice/src/main/java/com/expedia/adaptivealerting/modelservice/util/BeanUtil.java
```diff
@@ -0,0 +1,41 @@
...
+import org.springframework.beans.BeansException;
+import org.springframework.context.ApplicationContext;
+import org.springframework.context.ApplicationContextAware;
+import org.springframework.stereotype.Service;
+
+/**
+ * <p>
+ * We cannot inject any Spring-managed bean in any of the provider classes because these are instantiated by JPA before Spring injects anything into it.
+ * This is a utility class to fetch any bean according to our requirements.
+ * </p>
```
williewheeler: I think I asked this before--can you help me understand why JPA is involved at all? What exactly is JPA instantiating here?

modelservice/src/main/java/com/expedia/adaptivealerting/modelservice/providers/graphite/Tags.java
karanshah03: yeah. good point. fixed it.

modelservice/src/main/java/com/expedia/adaptivealerting/modelservice/providers/graphite/GraphiteResult.java
karanshah03: This is fixed now.

modelservice/src/main/java/com/expedia/adaptivealerting/modelservice/providers/graphite/GraphiteMetricSource.java
karanshah03: yeah. configurable isn't even required here. I was trying to do something with that earlier but forgot to removed it. I have removed it now.

modelservice/src/main/java/com/expedia/adaptivealerting/modelservice/util/BeanUtil.java
karanshah03: Thanks for calling that out. This is fixed.

modelservice/src/main/java/com/expedia/adaptivealerting/modelservice/web/AnomalyController.java
karanshah03: done

modelservice/src/test/java/com/expedia/adaptivealerting/modelservice/providers/graphite/GraphiteMetricSourceTest.java
karanshah03: yeah. I like the pattern you suggested and have updated it.

modelservice/src/test/java/com/expedia/adaptivealerting/modelservice/providers/graphite/GraphiteMetricSourceTest.java
karanshah03: fixed

modelservice/src/main/java/com/expedia/adaptivealerting/modelservice/providers/graphite/GraphiteMetricSource.java
```diff
@@ -0,0 +1,63 @@
...
+public class GraphiteMetricSource implements MetricSource {
+
+    private RestTemplate restTemplate = new RestTemplate();
```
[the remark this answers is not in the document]
karanshah03: yeah. So the thing is we cannot inject any Spring-managed bean in any of the provider classes because these are instantiated before Spring injects anything into it and returns null if I just do `@Autowired RestTemplate restTemplate`

modelservice/src/main/java/com/expedia/adaptivealerting/modelservice/providers/graphite/Tags.java
williewheeler: Wouldn't we just remove the `Tags` class?

PR #480, opened 2019-04-18
bug fix for model service

karanshah03:
Updating how we define metricSources list. `List<? extends MetricSource> metricSources` results in an injection error. We will have to cast it wherever it's required.


karanshah03:
Not required anymore. https://github.com/ExpediaDotCom/adaptive-alerting/pull/481 took care of this.

PR #487, opened 2019-05-08
Enhacement/detectors elastic search

karanshah03:


modelservice/src/main/java/com/expedia/adaptivealerting/modelservice/repo/impl/ElasticSearchDetectorRepositoryImpl.java
```diff
@@ -0,0 +1,91 @@
...
+    private static final int RESULTS_SIZE = 500;
+
+    @Autowired
+    private ElasticsearchTemplate elasticsearchTemplate;
+
+    @Autowired
+    private Client elasticSearchClient;
+
+    @Override
+    public void toggleDetector(ElasticSearchDetector detector, Boolean enabled) {
```
hridyeshpant: the name of this method looks confusion, it is either adding the detector or updating the detector but saying this method `toggleDetector` does not give any clue about the action.

modelservice/src/main/java/com/expedia/adaptivealerting/modelservice/repo/impl/ElasticSearchDetectorRepositoryImpl.java
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    @Override
+    public List<ElasticSearchDetector> getLastUpdatedDetectors(String fromDate, String toDate) {
+
+        val matchDocumentsWithinRange = buildQueryBuilder(fromDate, toDate);
+        val hits = getElasticSearchResponse(matchDocumentsWithinRange).getHits();
+        List<ElasticSearchDetector> elasticSearchDetectors = new ArrayList<>();
+
+        for (val hit : hits) {
```
hridyeshpant: can we do like 
hits.forEach(elasticSearchDetectors.add(getElasticSearchDetector(hit.getSource()))); ?

modelservice/src/main/java/com/expedia/adaptivealerting/modelservice/repo/impl/ElasticSearchDetectorRepositoryImpl.java
```diff
@@ -0,0 +1,91 @@
...
+        return elasticSearchClient.prepareSearch()
+                .setSize(RESULTS_SIZE)
+                .setQuery(queryBuilder)
+                .execute().actionGet();
+    }
+
+    private ElasticSearchDetector getElasticSearchDetector(Map<String, Object> source) {
+
+        val elasticSearchDetector = new ElasticSearchDetector();
+        elasticSearchDetector.setId((String) source.get("id"));
```
hridyeshpant: Can we use  Builder Design Pattern to set values here?

modelservice/src/test/java/com/expedia/adaptivealerting/modelservice/web/DetectorControllerTest.java
```diff
@@ -0,0 +1,50 @@
...
+    private ElasticSearchDetectorRepository repo;
+
+    @Before
+    public void setUp() {
+        MockitoAnnotations.initMocks(this);
+    }
+
+    @Test
+    public void testToggleDetector() {
+        elasticSearchService.toggleDetector("uuid", true);
```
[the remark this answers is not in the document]
karanshah03: toggle detector doesn't return anything. it has a void return type

modelservice/src/main/java/com/expedia/adaptivealerting/modelservice/repo/impl/ElasticSearchDetectorRepositoryImpl.java
karanshah03: i am just using the existing method names so that we can introduce the change in a non-breaking manner.

modelservice/src/main/resources/application.yml
```diff
@@ -33,11 +33,15 @@ management:
       tomcat: false
       system: false
       process: false
+elasticsearch:
+  clusterName: elasticsearch_kashah
```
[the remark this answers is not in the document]
karanshah03: will fix it

modelservice/src/main/java/com/expedia/adaptivealerting/modelservice/repo/impl/ElasticSearchDetectorRepositoryImpl.java
karanshah03: this is for each loop only and it's preferable over normal for loop as it gets rid of the clutter and avoids errors by hiding the iterator or index variable completely.

modelservice/src/main/java/com/expedia/adaptivealerting/modelservice/ElasticSearchConfig.java
```diff
@@ -0,0 +1,56 @@
...
+    @Bean
+    @SneakyThrows
+    public Client client() {
+        Settings elasticsearchSettings = Settings.builder()
+                .put("client.transport.sniff", true)
+                .put("cluster.name", settings.getClusterName()).build();
+
+        TransportClient client = new PreBuiltTransportClient(elasticsearchSettings);
+        client.addTransportAddress(new InetSocketTransportAddress(InetAddress.getByName(settings.getHostName()), settings.getPort()));
+        return client;
```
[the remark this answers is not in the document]
karanshah03: yup. that shouldn't cause any problem.

modelservice/src/test/java/com/expedia/adaptivealerting/modelservice/web/DetectorControllerTest.java
```diff
@@ -0,0 +1,50 @@
...
+
+    @Mock
+    private ElasticSearchService elasticSearchService;
+
+    @Mock
+    private ElasticSearchDetectorRepository repo;
+
+    @Before
+    public void setUp() {
+        MockitoAnnotations.initMocks(this);
```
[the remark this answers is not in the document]
karanshah03: yeah. good catch. we don't need it. will remove.

modelservice/src/main/java/com/expedia/adaptivealerting/modelservice/repo/impl/ElasticSearchDetectorRepositoryImpl.java
karanshah03: actually, let me just change the function names since now we are defining an explicit controller and not using the repo function.

karanshah03: actually toggle detector sounds like a right word for this. https://stackoverflow.com/questions/2544435/whats-the-word-for-enable-disable

karanshah03:
raising a new one for this.

PR #497, opened 2019-05-17
fixing terraform bug

shsethi:


williewheeler:
Thanks @karanshah03 for the assist.

PR #500, opened 2019-05-22
Enhancement/es detectors

karanshah03:


modelservice/src/main/java/com/expedia/adaptivealerting/modelservice/web/DetectorMappingController.java
```diff
@@ -41,7 +41,7 @@
     private DetectorMappingService detectorMappingService;
 
     @RequestMapping(produces = "application/json", method = RequestMethod.GET)
-    public DetectorMapping getDetectorMapping(@RequestParam String id) {
+    public ElasticsearchDetectorMapping getDetectorMapping(@RequestParam String id) {
```
williewheeler: Why does the controller know about the backend data store? Shouldn't Elasticsearch be invisible to the controller?

modelservice/src/main/java/com/expedia/adaptivealerting/modelservice/web/DetectorController.java
```diff
@@ -0,0 +1,83 @@
...
+import org.springframework.web.bind.annotation.RequestBody;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RequestParam;
+import org.springframework.web.bind.annotation.ResponseStatus;
+import org.springframework.web.bind.annotation.RestController;
+
+import java.util.List;
+
+@RestController
+@RequestMapping(path = "/api/elasticSearchDetectors")
```
williewheeler: The controller shouldn't expose the backend data store in the path.

modelservice/src/main/java/com/expedia/adaptivealerting/modelservice/web/DetectorController.java
```diff
@@ -0,0 +1,83 @@
...
+        detectorService.updateDetector(uuid, elasticsearchDetector);
+    }
+
+    @DeleteMapping
+    @ResponseStatus(HttpStatus.OK)
+    public void deleteDetector(@RequestParam String uuid) {
+        detectorService.deleteDetector(uuid);
+    }
+
+    @GetMapping(path = "/findByUuid", produces = "application/json")
```
williewheeler: Use `/search/findByUuid` for the search-related paths, since we are trying to pre-align it with SD Elasticsearch. Similarly for `/search/findByCreatedBy`

modelservice/src/main/java/com/expedia/adaptivealerting/modelservice/web/DetectorMappingController.java
karanshah03: We still have detector mapping entity for mysql and here I have just renamed detector mapping entity to ElasticsearchDetectorMapping and put it inside entity package. Once we get rid of mysql, we can fix it.

modelservice/src/main/java/com/expedia/adaptivealerting/modelservice/web/DetectorController.java
karanshah03: yeah. I used this path name to distinguish with the current detector path we have which is being used for mysql.

modelservice/src/main/java/com/expedia/adaptivealerting/modelservice/web/DetectorController.java
karanshah03: sure thing. will fix it.

PR #510, opened 2019-06-12
Enhancement/breakout

williewheeler:


anomdetect/src/main/java/com/expedia/adaptivealerting/anomdetect/breakout/Edmx.java
```diff
@@ -0,0 +1,94 @@
...
+     *
+     * @param data  Time series data
+     * @param delta Minimum sample size for computing a median. This applies to both the left and the right medians.
+     * @return EDM-X breakout result
+     */
+    public static BreakoutResult edmx(double[] data, int delta) {
+        val scaledData = unitScale(data);
+        val n = scaledData.length;
+
+        double bestStat = Double.MIN_VALUE;
```
williewheeler: Hm, I think this is wrong. With this, we will always have a breakout. Will look into this a bit more.

PR #511, opened 2019-06-13
Refactor/anomdetect

williewheeler:


anomdetect/src/test/java/com/expedia/adaptivealerting/anomdetect/outlier/forecast/point/HoltWintersForecasterTest.java
```diff
@@ -30,11 +32,6 @@
 import java.util.List;
 import java.util.ListIterator;
 
-import static com.expedia.adaptivealerting.anomdetect.forecast.point.holtwinters.HoltWintersAustouristsTestHelper.AUSTOURISTS_ADD_DATA;
```
mattcallanan: Curious if undoing static imports was an intentional stylistic change or an accidental by-product of the refactoring?
If the former, it seems inconsistently applied for `assertArrayEquals` in this file and `HoltWinters*Test.java` files.

PR #517, opened 2019-06-19
added bug fix

karanshah03:
added bugfix if UUID is not provided as part of the body while updating a detector


modelservice/src/main/java/com/expedia/adaptivealerting/modelservice/repo/impl/DetectorRepositoryImpl.java
```diff
@@ -104,6 +103,8 @@ public void updateDetector(String uuid, Detector detector) {
                 if (name.equals("lastUpdateTimestamp")) {
```
williewheeler: Please reverse the order to “lastUpdateTimestamp”.equals(name) (similarly below) to avoid possible NPEs. This is a general habit I recommend, not limited to this PR.

PR #533, opened 2019-07-03
Cleanup/removed mysql

karanshah03:


adriancole:
Why was mysql here before and why is it being nuked? I have no problem with this, just curious

williewheeler:
@adriancole It was our data store originally. But one of the features that we want to support is a reverse lookup--specifically, when a metric point comes in, can we find the "subscribed" detectors (i.e. the detectors we should apply to that). That reverse lookup is a non-trivial search and we found that Elasticsearch has a product called Percolator that can handle that. We saw an opportunity to consolidate the databases too since most of what we store is actually detector documents anyway, and all things being equal I'd rather have one DB than two.

Let me know your thoughts. I'm going to merge this but happy to take feedback.

williewheeler:
@adriancole https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-percolate-query.html

Idea is that the metric-detector mappings are specified by query on metric tags. So when the metric point (document) comes in, we want to find the matching detectors (queries). The normal lookup is given a query, find the matching documents.

adriancole:
very interesting. thanks!

PR #544, opened 2019-07-24
Bug fix/sast scan

karanshah03:
Bugfix for the vulnerability pointed out by SAST scan.
1) Unreleased Stream
2) Path Manipulation
3) Access Specifier Manipulation
4) Null Dereference


williewheeler:
@karanshah03 Can you include notes in the code indicating that the changes in question remediate the issues uncovered by the SAST scan? I want to avoid going in later changing things without knowing why they are the way they are.

karanshah03:
ok sure. will do

PR #554, opened 2019-09-11
Pre-create index mapping for detector index

swfortier:
Adds new class to pre-create index mappings for detectors elasticsearch index if it does not exist.
This is required to spin up a working docker-compose instance of this project from scratch.

This adds additional index name and doctype parameters to the application.yml for the modelservice. If the new parameters are not present the index creation process is not started to remain compatible with existing configs.


williewheeler:
Please implement the unit test for this PR.

PR #566, opened 2019-09-17
Feature/add graphite source metric functions aggregator

dinilatgit:


metrics/src/main/java/com/expedia/adaptivealerting/metrics/functions/MetricFunctionsApp.java
```diff
@@ -0,0 +1,61 @@
...
+
+    private static final String METRIC_SOURCE_SINK = "metric-source-sink";
+    private static final String APP_ID = "aa-metric-functions";
+    private static final String AGGREGATOR_PRODUCER = "aggregator-producer";
+    private static final String INPUT_FUNCTIONS_FILENAME ="functions.txt";
+    private final static String INPUT_FILE_PATH = "/config/";
+    private static final int corePoolSize = 5;
+
+    public static void main(String[] args) {
+        val config = new TypesafeConfigLoader(APP_ID).loadMergedConfig();
```
hridyeshpant: i think `val` stuff is getting the issue in SAST scans but not 100% sure

metrics/src/main/java/com/expedia/adaptivealerting/metrics/functions/MetricFunctionsApp.java
```diff
@@ -0,0 +1,61 @@
...
+        val metricSourceSinkConfig = config.getConfig(METRIC_SOURCE_SINK);
+        val aggregatorProducerConfig = config.getConfig(AGGREGATOR_PRODUCER);
+        val aggregatorProducerProps = ConfigUtil.toProducerConfig(aggregatorProducerConfig);
+        Producer<String, MetricData> aggregatorProducer = new KafkaProducer<String, MetricData>(aggregatorProducerProps);
```
hridyeshpant: new KafkaProducer<>

metrics/src/test/java/com/expedia/adaptivealerting/metrics/functions/source/GraphiteQueryResultTest.java
```diff
@@ -0,0 +1,31 @@
...
+import org.junit.Test;
+import java.util.Iterator;
+import java.util.Map;
+
+import static org.junit.Assert.assertEquals;
+
+public class GraphiteQueryResultTest {
+
+    @Test
+    public void testgetGraphiteQueryResultFromJson(){
```
hridyeshpant: what if sampleJsonGraphite value is is empty?

metrics/src/main/java/com/expedia/adaptivealerting/metrics/functions/source/GraphiteQueryResult.java
```diff
@@ -0,0 +1,72 @@
...
+        Datapoint = new Datapoint(0.0, 0);
+        Target = "";
+        Tags = new HashMap<>();
+    }
+
+    public void getGraphiteQueryResultFromJson(String jsonGraphiteOutput) {
+        JSONTokener graphiteQuery = new JSONTokener(jsonGraphiteOutput);
+        JSONArray graphiteResultJsonArray = new JSONArray(graphiteQuery);
+        JSONObject graphiteResultJsonObject = graphiteResultJsonArray.getJSONObject(0);
+        String[] datapointString = graphiteResultJsonObject.getJSONArray(GRAPHITE_RESULT_DATAPOINTS_KEY).get(0).toString().split(",");
```
hridyeshpant: i think better to handle cases where get(0) or toString().split(",") failing, similar in test case as well

metrics/src/main/java/com/expedia/adaptivealerting/metrics/functions/MetricFunctionsApp.java
dinilatgit: the same is used, expanded default.

metrics/src/main/java/com/expedia/adaptivealerting/metrics/functions/MetricFunctionsApp.java
dinilatgit: val is recommended use.

metrics/src/test/java/com/expedia/adaptivealerting/metrics/functions/source/GraphiteQueryResultTest.java
dinilatgit: added tests

metrics/src/main/java/com/expedia/adaptivealerting/metrics/functions/source/GraphiteQueryResult.java
dinilatgit: added checks

metrics/src/main/java/com/expedia/adaptivealerting/metrics/functions/source/GraphiteQueryResult.java
```diff
@@ -0,0 +1,72 @@
...
+     *  [ {"datapoints": [ [value1, timestamp1], [value2, timestamp2] ..],
+     *  "target": "time-series-key-identier", "tags" : "<tags from graphite function used in the render api>"} ]
+     */
+
+    private final String GRAPHITE_RESULT_DATAPOINTS_KEY = "datapoints";
+    private final String GRAPHITE_RESULT_TARGET_KEY = "target";
+    private final String GRAPHITE_RESULT_TAGS_KEY = "tags";
+
+    @Getter
+    private Datapoint Datapoint;
```
karanshah03: The variables should follow camel case. should be datapoint here.

metrics/src/main/java/com/expedia/adaptivealerting/metrics/functions/source/GraphiteQueryResult.java
```diff
@@ -0,0 +1,72 @@
...
+
+    @Getter
+    private String Target;
```
karanshah03: same here

metrics/src/main/java/com/expedia/adaptivealerting/metrics/functions/source/GraphiteQueryResult.java
```diff
@@ -0,0 +1,72 @@
...
+
+    @Getter
+    private HashMap<String, String> Tags;
```
karanshah03: same here

metrics/src/main/java/com/expedia/adaptivealerting/metrics/functions/source/MetricFunctionsReader.java
```diff
@@ -0,0 +1,49 @@
...
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.nio.charset.StandardCharsets;
+import lombok.extern.slf4j.Slf4j;
+
+@Slf4j
+public class MetricFunctionsReader {
+
+    public static List<MetricFunctionsSpec> readFromInputFile(String InputFilename) {
+        List<MetricFunctionsSpec> metricFunctionSpecList = new ArrayList<MetricFunctionsSpec>();
```
karanshah03: MetricFunctionsSpec in  `new ArrayList<MetricFunctionsSpec>()` isredundant here.

metrics/src/main/java/com/expedia/adaptivealerting/metrics/functions/MetricFunctionsApp.java
```diff
@@ -0,0 +1,61 @@
...
+public class MetricFunctionsApp {
+
+    private static final String METRIC_SOURCE_SINK = "metric-source-sink";
+    private static final String APP_ID = "aa-metric-functions";
+    private static final String AGGREGATOR_PRODUCER = "aggregator-producer";
+    private static final String INPUT_FUNCTIONS_FILENAME ="functions.txt";
+    private final static String INPUT_FILE_PATH = "/config/";
+    private static final int corePoolSize = 5;
+
+    public static void main(String[] args) {
```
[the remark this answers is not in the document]
dinilatgit: The input to the module is not kafka, so this cannot be done.

metrics/src/main/java/com/expedia/adaptivealerting/metrics/functions/MetricFunctionsTask.java
```diff
@@ -0,0 +1,75 @@
...
+        this.metricFunctionsSpec = metricFunctionsSpec;
+        this.aggregateProducer = aggregateProducer;
+        this.metricSourceSinkConfig = metricSourceSinkConfig;
+    }
+
+    @SuppressWarnings("unchecked")
+    public void run() {
+        HttpClientWrapper metricFunctionHttpClient = new HttpClientWrapper();
+        /* For now source supported is graphite alone */
+        GraphiteQueryResult graphiteQueryResult = new GraphiteQueryResult();
```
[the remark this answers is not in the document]
dinilatgit: Implemented this.

metrics/src/main/java/com/expedia/adaptivealerting/metrics/functions/util/ConstructSourceURI.java
```diff
@@ -0,0 +1,44 @@
...
+        String URI = metricSourceSinkConfig.getString(GRAPHITE_HOST_KEY)
+                + GRAPHITE_RENDER_API_DEFINITION
+                + metricFunctionsSpec.getFunction()
+                + URI_CONSTRUCT_AMPERSAND
+                + GRAPHITE_FROM_TIME_PARAM_STRING
+                + metricFunctionsSpec.getIntervalInSecs() + GRAPHITE_TIME_UNIT
+                + URI_CONSTRUCT_AMPERSAND
+                + GRAPHITE_TO_TIME_PARAM_STRING
+                + URI_CONSTRUCT_AMPERSAND
+                + GRAPHITE_FORMAT_PARAM_STRING;
```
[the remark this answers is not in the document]
dinilatgit: simplified this.

metrics/src/main/java/com/expedia/adaptivealerting/metrics/functions/source/GraphiteQueryResult.java
```diff
@@ -0,0 +1,72 @@
+/*
+ * Copyright 2018-2019 Expedia Group, Inc.
```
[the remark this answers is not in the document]
dinilatgit: Done

metrics/src/main/java/com/expedia/adaptivealerting/metrics/functions/source/GraphiteQueryResult.java
dinilatgit: Fixed

metrics/src/main/java/com/expedia/adaptivealerting/metrics/functions/source/GraphiteQueryResult.java
dinilatgit: fixed

metrics/src/main/java/com/expedia/adaptivealerting/metrics/functions/source/GraphiteQueryResult.java
dinilatgit: fixed

metrics/src/main/java/com/expedia/adaptivealerting/metrics/functions/source/MetricFunctionsReader.java
dinilatgit: fixed

williewheeler:
Please add unit tests -- 70.14% is significantly below target, and this creates an over 2.54% drop in overall coverage

metrics/pom.xml
```diff
@@ -44,10 +49,19 @@
...
+            <artifactId>adaptive-alerting-kafka</artifactId>
+        </dependency>
         <dependency>
             <groupId>com.fasterxml.jackson.core</groupId>
             <artifactId>jackson-core</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.apache.httpcomponents</groupId>
+            <artifactId>fluent-hc</artifactId>
+            <version>${apache.httpcomponents.version}</version>
```
williewheeler: See above - manage dependencies from top POM please.

metrics/pom.xml
```diff
@@ -56,12 +70,20 @@
...
-            <scope>test</scope>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-databind</artifactId>
+            <version>${jackson.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.json</groupId>
+            <artifactId>json</artifactId>
+            <version>${json.version}</version>
+        </dependency>
```
williewheeler: I'm not familiar with this `org.json` library, but looking at the code it looks like it duplicates what we are already doing with Jackson. If that's the case I would ask that you use Jackson instead as I would like to avoid a needless proliferation of dependencies.

metrics/src/main/java/com/expedia/adaptivealerting/metrics/functions/MetricFunctionsApp.java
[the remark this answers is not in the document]
williewheeler: @dinilatgit -- I don't follow your response to @karanshah03 's suggestion.

The idea is that there's a core logic to metric functions that has nothing to do with Kafka. So the input to your module would be independent of the messaging implementation. So couldn't you take the Kafka-specific part and move that to the Kafka module?

Another way to think of this is what happens when we want to add support for Kinesis? Do we have to reimplement your entire metrics xform code for Kinesis? It would be better if they could just implement a Kinesis wrapper for the core logic.

metrics/src/test/java/com/expedia/adaptivealerting/metrics/functions/source/graphite/GraphiteQueryResultTest.java
```diff
@@ -0,0 +1,81 @@
...
+
+import static org.junit.Assert.assertEquals;
+
+public class GraphiteQueryResultTest {
+
+    @Test
+    public void testgetGraphiteQueryResultFromJson(){
+       val sampleJsonGraphite = "[{" + "\"datapoints\": [[12.0, 1568255056]], " +
+               "\"target\": \"sumSeries(a.b.c)\", " +
+               "\"tags\":" + "{" + "\"aggregatedBy\": \"sum\", \"name\": \"sumSeries(a.b.c)\"" + "}}]";
```
williewheeler: Consider creating an actual class to represent this Graphite data type, and then using Jackson to serialize it to text. This is easier to manage than hand-crafting strings with all the escaped quotes.

metrics/src/test/java/com/expedia/adaptivealerting/metrics/functions/source/graphite/GraphiteQueryResultTest.java
```diff
@@ -0,0 +1,81 @@
...
+import org.junit.Test;
+import java.util.Iterator;
+import java.util.Map;
+
+import static org.junit.Assert.assertEquals;
+
+public class GraphiteQueryResultTest {
+
+    @Test
+    public void testgetGraphiteQueryResultFromJson(){
```
williewheeler: Please use the standard Java idiom for method names. Meaning prefer `testGet` to `testget`. This applies throughout.

metrics/src/test/java/com/expedia/adaptivealerting/metrics/functions/service/MetricQueryServiceTest.java
```diff
@@ -0,0 +1,168 @@
...
+    @Before
+    public void initMetricSourceSinkConfigSetup() throws Exception {
+        MockitoAnnotations.initMocks(this);
+        val config = new TypesafeConfigLoader(APP_ID).loadMergedConfig();
+        metricSourceSinkConfig = config.getConfig(METRIC_SOURCE_SINK);
+        val functionsInputFile = "/config/functions-test.txt";
+        metricFunctionsSpec = MetricFunctionsReader.readFromInputFile(functionsInputFile).get(0);
+        val sampleJsonGraphite = "[{" + "\"datapoints\": [[12.0, 1568255056]], " +
+                "\"target\": \"sumSeries(a.b.c)\", " +
+                "\"tags\":" + "{" + "\"aggregatedBy\": \"sum\", \"name\": \"sumSeries(a.b.c)\"" + "}}]";
```
williewheeler: See comment below - consider using explicit class and serializing to avoid having to hand-craft strings.

metrics/src/test/java/com/expedia/adaptivealerting/metrics/functions/service/MetricQueryServiceTest.java
```diff
@@ -0,0 +1,168 @@
...
+        when(httpClientWrapper.get(uri)).thenReturn(mockGraphiteResult);
+        metricQueryService = new MetricQueryService(httpClientWrapper);
+        initValidMetricDataSetup();
+        initDefaultMetricDataSetup();
+    }
+
+    public void initValidMetricDataSetup() {
+        val sampleJsonGraphite = "[{" + "\"datapoints\": [[12.0, 1568255056]], " +
+                "\"target\": \"sumSeries(a.b.c)\", " +
+                "\"tags\":" + "{" + "\"aggregatedBy\": \"sum\", \"name\": \"sumSeries(a.b.c)\"" + "}}]";
```
williewheeler: See comment below - consider using explicit class and serializing to avoid having to hand-craft strings.

metrics/src/main/java/com/expedia/adaptivealerting/metrics/functions/service/MetricQueryService.java
```diff
@@ -0,0 +1,98 @@
...
+@Slf4j
+public class MetricQueryService {
+    private HttpClientWrapper metricFunctionHttpClient;
+    private final String EMPTY_RESULT_FROM_SOURCE = "{}";
+    private final String METRICDATA_KEY_FOR_EMPTRY_DATA = "aggregator.producer.";
+    private final double METRICDATA_DEFAULT_VALUE = 0.0;
+    private final String METRIC_SOURCE_KEY = "metric-source";
+    /* For now source supported is graphite alone , non-graphite is just
+    *  a place holder for the common method
+    */
```
williewheeler: I would suggest removing the non-Graphite stuff altogether. If we were to add support for other backends, I'd probably expect this to happen via SPI plugins instead of just building out a single class like `MetricQueryService` (see https://en.wikipedia.org/wiki/Open%E2%80%93closed_principle). In any event we aren't currently doing this so it seems better just to keep things tight and remove placeholders for future ideas. Keeping things tight is itself a way to make things easier to extend and/or discard.

metrics/src/main/java/com/expedia/adaptivealerting/metrics/functions/service/MetricQueryService.java
```diff
@@ -0,0 +1,98 @@
...
+
+    /* common method for generic metric source */
+    public MetricData getMetricQueryResult(Config metricSourceSinkConfig, MetricFunctionsSpec metricFunctionsSpec) {
+        if (METRIC_SOURCE.contains(metricSourceSinkConfig.getString(METRIC_SOURCE_KEY))) {
+            return queryMetricSource(metricSourceSinkConfig, metricFunctionsSpec);
+        }
+        else {
+            return defaultMetricData(metricFunctionsSpec);
+        }
+    }
```
williewheeler: This if/else is an example of the sort of thing I wouldn't want to do, because it violates the OCP as I noted above. Please assume that this is Graphite-only right now, and that whatever Graphite-only code you're implementing will become an SPI plugin if we ever generalize, meaning that we still won't need support for non-Graphite inside this code.

williewheeler: In fact I'd suggest renaming this class to reflect the fact that it specifically targets Graphite: say `GraphiteMetricFunctions` or similar.

metrics/src/main/java/com/expedia/adaptivealerting/metrics/functions/source/MetricFunctionsReader.java
```diff
@@ -0,0 +1,49 @@
...
+            String metricFunctionSpecString;
+            while ((metricFunctionSpecString = bufferedReader.readLine()) != null) {
+                ObjectMapper objectMapper = new ObjectMapper();
+                metricFunctionSpecList.add(objectMapper.readValue(metricFunctionSpecString, MetricFunctionsSpec.class));
+            }
+        }
+        catch(Exception e) {
+            log.error("Exception while reading input functions'" +
+                    "definition", e);
+        }
```
williewheeler: After the exception, this method simply returns an empty list. Is that really the right behavior here? I question that. If this read fails, that seems pretty fatal to me--none of the aggregations will work, and this systems depends on those working.

williewheeler: Also, small nit -- if you break your strings onto multiple lines, pay attention to whitespace. We don't want the message to be

`Exception while reading input functions'definition`

metrics/src/main/java/com/expedia/adaptivealerting/metrics/functions/service/MetricQueryService.java
```diff
@@ -0,0 +1,98 @@
...
+
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.List;
+
+@Slf4j
+public class MetricQueryService {
+    private HttpClientWrapper metricFunctionHttpClient;
+    private final String EMPTY_RESULT_FROM_SOURCE = "{}";
+    private final String METRICDATA_KEY_FOR_EMPTRY_DATA = "aggregator.producer.";
```
williewheeler: `EMPTY`

metrics/src/main/java/com/expedia/adaptivealerting/metrics/functions/source/MetricFunctionsSpec.java
```diff
@@ -0,0 +1,34 @@
...
+ * limitations under the License.
+ */
+package com.expedia.adaptivealerting.metrics.functions.source;
+
+import java.util.HashMap;
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import lombok.Getter;
+
+@JsonIgnoreProperties
+public class MetricFunctionsSpec {
```
williewheeler: Apply the `@Data` annotation to the class instead of using `@Getter` here. This is about as `@Data` as it gets.

metrics/src/main/java/com/expedia/adaptivealerting/metrics/functions/source/graphite/Datapoint.java
```diff
@@ -0,0 +1,33 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.expedia.adaptivealerting.metrics.functions.source.graphite;
+
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+import lombok.Getter;
+
+@JsonSerialize
+public class Datapoint {
```
williewheeler: Apply `@Data` to the class rather than `@Getter` to the individual fields.

metrics/src/main/java/com/expedia/adaptivealerting/metrics/functions/source/graphite/GraphiteQueryResult.java
```diff
@@ -0,0 +1,85 @@
...
+
+import lombok.Getter;
+import org.json.JSONArray;
+import org.json.JSONObject;
+import org.json.JSONTokener;
+
+import java.util.HashMap;
+import java.util.Iterator;
+
+public class GraphiteQueryResult {
```
williewheeler: User `@Data`; see above.

metrics/src/main/java/com/expedia/adaptivealerting/metrics/functions/MetricFunctionsApp.java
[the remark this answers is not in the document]
dinilatgit: created KafkaMetricFunctions for the decoupling request.

metrics/pom.xml
```diff
@@ -33,6 +33,11 @@
     <name>${project.groupId}:${project.artifactId}</name>
     <description>Metrics processing.</description>
 
+    <properties>
+        <jackson.version>2.9.8</jackson.version>
+        <json.version>20180813</json.version>
+    </properties>
```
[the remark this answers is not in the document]
dinilatgit: Fixed.

metrics/pom.xml
dinilatgit: Fixed.

metrics/pom.xml
dinilatgit: The json returned does not have perfect json elements - datapoint. Jackson was tried but did not work. This library needs to be used for processing.

metrics/src/test/java/com/expedia/adaptivealerting/metrics/functions/source/graphite/GraphiteQueryResultTest.java
dinilatgit: Added changes to create test String from Object and not hand crafted string.

metrics/src/main/java/com/expedia/adaptivealerting/metrics/functions/source/MetricFunctionsReader.java
dinilatgit: Handling runtime exception in main for empty list.

metrics/src/main/java/com/expedia/adaptivealerting/metrics/functions/service/MetricQueryService.java
dinilatgit: Fixed.

metrics/src/main/java/com/expedia/adaptivealerting/metrics/functions/source/MetricFunctionsSpec.java
dinilatgit: Fixed.

metrics/src/main/java/com/expedia/adaptivealerting/metrics/functions/source/graphite/Datapoint.java
dinilatgit: Fixed.

metrics/src/main/java/com/expedia/adaptivealerting/metrics/functions/source/graphite/GraphiteQueryResult.java
dinilatgit: Fixed.

metrics/src/main/java/com/expedia/adaptivealerting/metrics/functions/MetricFunctionsApp.java
dinilatgit: Fixed.

metrics/pom.xml
williewheeler: Please remove the `version` definitions from this file. They are automatically inherited.

metrics/pom.xml
williewheeler: I need help understanding what won't work.

metrics/src/main/java/com/expedia/adaptivealerting/metrics/functions/MetricFunctionsApp.java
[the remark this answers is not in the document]
williewheeler: Let's pair on this one.

metrics/src/main/java/com/expedia/adaptivealerting/metrics/functions/MetricFunctionsTask.java
```diff
@@ -0,0 +1,75 @@
...
+                log.error("Exception during reading from metric source", e);
+
+        }
+
+        try {
+            MetricData metricData = KafkaMetricFunctionsToAAMetrics.streamAggregateRecord(graphiteQueryResult, metricFunctionsSpec);
+            ProducerRecord aggregateProducerRecord = new ProducerRecord(
+                    metricSourceSinkConfig.getString(OUTPUT_TOPIC_KEY_STRING),
+                    metricData.getMetricDefinition().getKey(), metricData);
+            aggregateProducer.send(aggregateProducerRecord);
```
williewheeler: This task is presumably part of the core logic for calling Graphite, and as such shouldn't know anything about Kafka (or Kinesis or any other messaging). I believe what we want is for a Kafka wrapper to create a callback, pass it down, and then this task invokes the callback to send the transformed metrics.

metrics/src/main/java/com/expedia/adaptivealerting/metrics/functions/source/GraphiteQueryResult.java
```diff
@@ -0,0 +1,72 @@
...
+
+    public void getGraphiteQueryResultFromJson(String jsonGraphiteOutput) {
+        JSONTokener graphiteQuery = new JSONTokener(jsonGraphiteOutput);
+        JSONArray graphiteResultJsonArray = new JSONArray(graphiteQuery);
+        JSONObject graphiteResultJsonObject = graphiteResultJsonArray.getJSONObject(0);
+        String[] datapointString = graphiteResultJsonObject.getJSONArray(GRAPHITE_RESULT_DATAPOINTS_KEY).get(0).toString().split(",");
+        this.Datapoint = new Datapoint(Double.parseDouble(datapointString[0].substring(1)), Long.parseLong(datapointString[1].substring(0, datapointString[1].length() - 1 )));
+        this.Target = graphiteResultJsonObject.getString(GRAPHITE_RESULT_TARGET_KEY);
+        setTagsFromString(graphiteResultJsonObject.get(GRAPHITE_RESULT_TAGS_KEY).toString());
+    }
```
williewheeler: We should be able to do this from Jackson, and I would like to give that another shot before introducing a second JSON parsing dependency.

williewheeler: E.g. https://stackoverflow.com/questions/27330136/java-jackson-nested-arrays

metrics/src/main/java/com/expedia/adaptivealerting/metrics/functions/source/graphite/Datapoint.java
```diff
@@ -0,0 +1,32 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.expedia.adaptivealerting.metrics.functions.source.graphite;
+
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+import lombok.Data;
+
+@JsonSerialize
+@Data
```
williewheeler: Shouldn't need `@JsonSerialize` here. This is an extremely straightforward class and shouldn't require custom serialization.

metrics/src/main/java/com/expedia/adaptivealerting/metrics/functions/source/graphite/Datapoint.java
```diff
@@ -0,0 +1,32 @@
...
+public class Datapoint {
+    private double value;
+
+    private long timestamp;
+
+    public Datapoint(double value, long timestamp){
+        this.value = value;
+        this.timestamp = timestamp;
+    }
```
williewheeler: Note that you can use `@AllArgsConstructor` here. Not required but it's an option.

metrics/src/main/java/com/expedia/adaptivealerting/metrics/functions/source/graphite/GraphiteQueryResult.java
```diff
@@ -0,0 +1,83 @@
...
+        datapoint = new Datapoint(0.0, 0);
+        target = "";
+        tags = new HashMap<>();
+    }
+
+    public void getGraphiteQueryResultFromJson(String jsonGraphiteOutput) {
+        JSONTokener graphiteQuery = new JSONTokener(jsonGraphiteOutput);
+        JSONArray graphiteResultJsonArray = new JSONArray(graphiteQuery);
+        if (graphiteResultJsonArray.length() > 0) {
+            JSONObject graphiteResultJsonObject = graphiteResultJsonArray.getJSONObject(0);
```
williewheeler: This is the part I want to see if we can do using Jackson. Happy to work with you to see if we can make it work.

metrics/src/main/java/com/expedia/adaptivealerting/metrics/functions/MetricFunctionsApp.java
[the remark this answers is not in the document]
williewheeler: OK looks like we won't have time to sync today, so let me describe what I have in mind. There should be two general pieces here -- a core metric functions component and a Kafka wrapper. The Kafka wrapper should call the metric functions and not vice versa -- we want to be able to add other wrappers like Kinesis and so that means that the core metric functions can't know about its messaging environment impl.

This leaves open the question of how the metric functions comp is supposed to publish its metrics to Kafka if it doesn't know it's running in Kafka. To solve this, create a generic callback interface for publishing metrics. You can call it `TransformedMetricTopic` or similar. It should have a single method that allows you to publish the metric in question, and the method shouldn't know about Kafka.

Then when you initialize the Kafka wrapper, implement the callback interface and pass that into the core metric functions component. That gives the component a way to publish metrics to Kafka without having to know anything about Kafka.

metrics/src/main/java/com/expedia/adaptivealerting/metrics/functions/source/GraphiteQueryResult.java
dinilatgit: The inner class in our case is not an array of same object, nor can key value pair be used, but i can further review for other ways to try this with Jackson.

metrics/src/main/java/com/expedia/adaptivealerting/metrics/functions/MetricFunctionsTask.java
dinilatgit: Added interface, moved main to kafka module.

metrics/src/main/java/com/expedia/adaptivealerting/metrics/functions/source/graphite/Datapoint.java
dinilatgit: Fixed.

metrics/src/main/java/com/expedia/adaptivealerting/metrics/functions/source/graphite/Datapoint.java
dinilatgit: Fixed.

PR #567, opened 2019-09-17
Add Docker Compose

swfortier:
Docker-compose configuration for the pipeline along with a demo script to send sample metrics and retrieve anomalies.

### DIFF
--- a/docker-compose/README.md
+++ b/docker-compose/README.md
@@ -1,0 +1,25 @@
+# Testing the Adaptive Alerting pipeline with Docker Compose
+
+The Docker Compose setup allows you run the Adaptive Alerting components and test sample metrics against the project. By default the images are downloaded from Docker Hub, but they can be subsituted with your own builds or different versions from Docker Hub for testing purposes.
+
+#### 1. Build and run using Docker Compose
+```
+docker-compose up
+```
+
+#### 2. Send sample events to running environment
+Wait a few minutes for docker-compose to finish and all the services to come up before executing `send-sample-metrics.py`. If no anomalies are generated it could be because the sample detector has not yet been inserted.
+```
+python3 send-sample-metrics.py
+```
+
+**Note**: This Python script sends randomized sample metrics into the intake Kafka topic and show any anomalous values. Anomalies are determined by the sample constant threshold detector and detector mapping that is populated upon bringing up docker-compose. Requires: [msgpack](https://msgpack.org/index.html) and [kafka-python](https://kafka-python.readthedocs.io/en/master/) libraries.
+
+#### 3. Reset Docker Compose and remove data volumes
+```
+docker-compose down -v
+```
+
+#### Support Files executed by Elasticsearch container
+ - `/scripts/start-es.sh`: Kicks off `populate-es.sh` in the background and brings up Elasticsearch.
+ - `/scripts/populate-es.sh`: Pushes a sample detector and detector mapping to the modelservice API as soon as the API is available.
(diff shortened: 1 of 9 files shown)


docker-compose/send-sample-metrics.py
```diff
@@ -0,0 +1,114 @@
...
+
+
+def consume_from_kafka(topic):
+    output_count = 0
+    anomaly_count = 0
+    strong_values = []
+    weak_values = []
+    consumer = KafkaConsumer(topic,
+                            bootstrap_servers=kafka_boostrap_servers, 
+                            auto_offset_reset='earliest',
```
hridyeshpant: don't we want to use latest?

docker-compose/send-sample-metrics.py
```diff
@@ -0,0 +1,114 @@
...
+                            bootstrap_servers=kafka_boostrap_servers, 
+                            auto_offset_reset='earliest', 
+                            enable_auto_commit=False, 
+                            consumer_timeout_ms=5000)
+    
+    for message in consumer:
+        output_count+=1
+        anomaly = json.loads(message.value)
+        anomalyLevel = anomaly.get('anomalyResult',{}).get('anomalyLevel')
+        if anomalyLevel == 'WEAK':
```
hridyeshpant: anomaly_value = {}
if anomalyLevel == 'WEAK' or anomalyLevel == 'STRONG'
    anomaly_count+=1
     anomaly_value[anomalyLevel] = anomaly.get('metricData').get('value')
return anomaly_value


and use weak_anomaly_values = anomaly_value['WEAK']

docker-compose/docker-compose.yml
```diff
@@ -0,0 +1,146 @@
...
+#       distributed under the License is distributed on an "AS IS" BASIS,
+#       WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#       See the License for the specific language governing permissions and
+#       limitations under the License.
+#
+version: "2.4"
+services:
+
+  # AA Components
+  modelservice:
```
hridyeshpant: i see one issue here for startup Compose does not wait until a container is “ready”  only until it’s running.
please see https://docs.docker.com/compose/startup-order/

docker-compose/send-sample-metrics.py
swfortier: I can do latest, but I will need to run the consumer function in a separate thread since I have to be simultaneously reading from the topic while metrics are being sent in. Otherwise the topic will be empty.

docker-compose/docker-compose.yml
swfortier: I am using the docker compose 2.4 features "healthcheck" + "depends_on condition" to ensure containers start up when their dependencies are ready. Containers will not start until their dependencies report success on their healthcheck command.

docker-compose/README.md
```diff
@@ -0,0 +1,25 @@
+# Testing the Adaptive Alerting pipeline with Docker Compose
+
+The Docker Compose setup allows you run the Adaptive Alerting components and test sample metrics against the project. By default the images are downloaded from Docker Hub, but they can be subsituted with your own builds or different versions from Docker Hub for testing purposes.
```
williewheeler: s/subsituted/substituted

docker-compose/docker-compose.yml
williewheeler: I am fine with this for now, but we need to be able to move forward as Docker Compose releases new versions. In their docs they mention a `wait-for.sh` script that we can use.

swfortier: I do have a version 3 one that does work, but it relies on the modelservice dying and restarting until it works (usually takes 2 times), so I wasn't happy with it. The wait for it script might be the best approach for a v3 compose script unless we refactor the model service to be more graceful about Kafka not being available.

PR #575, opened 2019-09-24
getting rid of notifier

cmoroexpedia:


williewheeler:
Not sure whether we want to keep the notifier at this point. If we decide to remove it, it's easy to do.

PR #576, opened 2019-09-26
Added aquila trainer to train seasonal time series

karanshah03:


williewheeler:
Oh, interesting. I was not necessarily thinking that this would go under the main OSS repo but I like that idea.

Can we call this "STL" rather than Aquila, though? This is just the forecaster part, and that's basically a slightly fancier version of STL that we will probably break up. Trying to keep the main repo for standard algos, not custom stuff.

trainer/aquila/aquila.ipynb
```diff
@@ -0,0 +1,292 @@
+{
```
williewheeler: For the moment let's do `trainer/stl.ipynb`.

PR #577, opened 2019-09-27
Enhacement/metric profiler batch calls

karanshah03:
Query elastic search in batches to fetch metric profiling info depending on the elastic search latency. This would reduce the load on the elastic search which in turn would also reduce the lag which we may observe in any of the apps in the pipeline.


karanshah03:
I will add more tests

karanshah03:
We have postponed this work for the moment.

williewheeler:
We moved profiler to the python codebase.

PR #597, opened 2019-10-11
Bugfix/handle null datapoint from source

dinilatgit:


metrics/src/main/java/com/expedia/adaptivealerting/metrics/functions/service/graphite/GraphiteQueryService.java
```diff
@@ -60,7 +60,8 @@ private String queryGraphiteSource(Config metricSourceSinkConfig, MetricFunction
     private MetricData graphiteMetricData(Config metricSourceSinkConfig, MetricFunctionsSpec metricFunctionsSpec) {
         String metricQueryResult = queryGraphiteSource(metricSourceSinkConfig, metricFunctionsSpec);
         GraphiteQueryResult graphiteQueryResult = new GraphiteQueryResult();
-        if (!metricQueryResult.equals(EMPTY_RESULT_FROM_SOURCE)) {
+        if (!(metricQueryResult.equals(EMPTY_RESULT_FROM_SOURCE) ||
```
williewheeler: To avoid `NullPointerException`, prefer

```
if (!(EMPTY_RESULT_FROM_SOURCE.equals(metricQueryResult) ||
```

metrics/src/test/java/com/expedia/adaptivealerting/metrics/functions/service/graphite/GraphiteQueryServiceTest.java
```diff
@@ -169,4 +169,37 @@ public void testGetMetricQueryResultEmpty() {
...
+        sampleJsonGraphiteResult.put("target", "sumSeries(a.b.c)");
+        sampleJsonGraphiteResult.put("tags", testTags);
+        sampleJsonGraphite.put(0, sampleJsonGraphiteResult);
+        Content mockGraphiteResult = new Content(sampleJsonGraphite.toString().getBytes(), ContentType.APPLICATION_JSON);
+        try {
+            when(httpClientWrapper.get(uri)).thenReturn(mockGraphiteResult);
+        }
+        catch (Exception e) {
+            log.error("{}", e);
+        }
```
williewheeler: After logging this exception, control continues flowing past the try/catch. Is that really the desired behavior?

I understand that it will not actually throw an exception here. But what I would suggest is just remove the try/catch and make the test method itself throw an `Exception`. That way if you get an unexpected exception, the test fails.

metrics/src/main/java/com/expedia/adaptivealerting/metrics/functions/service/graphite/GraphiteQueryService.java
dinilatgit: Fixed.

metrics/src/test/java/com/expedia/adaptivealerting/metrics/functions/service/graphite/GraphiteQueryServiceTest.java
dinilatgit: Updated.

PR #599, opened 2019-10-17
Updated DetectorDocument schema, and replaced legacy Detector class

williewheeler:
Regarding DetectorDocument, I updated the schema with a meta object for
audit-related fields, and harmonized the names. I left the old fields in
place for now so we can migrate over, but I will remove them after we do.

I also replaced the Detector class in the Model Service with the
DetectorDocument, as it's exactly the same thing, and it doesn't make
sense to migrate two equivalent classes.


modelservice/src/main/java/com/expedia/adaptivealerting/modelservice/repo/impl/DetectorRepositoryImpl.java
```diff
@@ -63,20 +62,19 @@
     private ElasticsearchUtil elasticsearchUtil;
 
     @Override
-    public String createDetector(Detector detector) {
-        String uuid = UUID.randomUUID().toString();
-        if (detector.getUuid() != null) {
-            uuid = detector.getUuid();
+    public String createDetector(DetectorDocument document) {
+        UUID uuid = UUID.randomUUID();
+        if (document.getUuid() != null) {
```
hridyeshpant: can we do something like this, in that case `UUID.randomUUID()` will only get called if condition is false
UUID uuid = document.getUuid()
uuid = uuid != null ? uuid : UUID.randomUUID()

williewheeler: I think what I would rather do is say that the UUID is required to be null coming in, and then check that coming in. That way it is understood that only the repo assigns UUIDs. Otherwise we will see client code generating UUID conflicts.

Issue #602, opened 2019-10-21
create checkstyle for java files


bibinss:
to start with, create following checks
1 - every java file should have a licence comment
2 - no java file should have unused imports
3 - every line should follow 4 level/space indentation.

PR #603, opened 2019-10-22
Created checkstyle for java files

haram-dev:
#602 I've added maven checkstyle for java files.
Although I've confirmed that checking styles of the files go well according to the requirements, I might have made some mistakes because I'm almost new to open source.
Any comments and suggestions are welcome.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -137,6 +137,9 @@
         <maven.source.plugin.version>3.0.1</maven.source.plugin.version>
         <maven.surefire.plugin.version>2.22.1</maven.surefire.plugin.version>
         <nexus.staging.maven.plugin.version>1.6.8</nexus.staging.maven.plugin.version>
+        <maven.site.plugin.version>3.7.1</maven.site.plugin.version>
+        <maven.project.info.reports.plugin.version>3.0.0</maven.project.info.reports.plugin.version>
+        <maven.checkstyle.plugin.version>3.0.0</maven.checkstyle.plugin.version>
     </properties>
 
     <dependencyManagement>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -751,6 +754,29 @@
...
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-project-info-reports-plugin</artifactId>
+                <version>${maven.project.info.reports.plugin.version}</version>
+            </plugin>
         </plugins>
     </build>
+
+    <reporting>
+        <plugins>
```
bibinss: we need to move this under build->plugins
also add a goal to the plugin
```
        <plugin>
                <groupId>org.apache.maven.plugins</groupId>
                <artifactId>maven-checkstyle-plugin</artifactId>
                <version>${maven.checkstyle.plugin.version}</version>
                <configuration>
                    <configLocation>checkstyle.xml</configLocation>
                </configuration>
                <executions>
                    <execution>
                        <goals>
                            <goal>check</goal>
                        </goals>
                    </execution>
                </executions>
            </plugin>```

File: pom.xml [hunk 2]
```diff
@@ -751,6 +754,29 @@
...
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-site-plugin</artifactId>
+                <version>${maven.site.plugin.version}</version>
+            </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-project-info-reports-plugin</artifactId>
+                <version>${maven.project.info.reports.plugin.version}</version>
+            </plugin>
```
bibinss: not sure we really need these plugins. why are these added?

pom.xml
bibinss: once you add this, you will see some checkstyle breakages and you can fix them also as part of this PR

bibinss:
@haram-dev will you be able to work on my suggestions here^^ ?

bibinss:
since I don't see a response/update here and since another dev has raised a new PR for this, I am going ahead with the new PR.

williewheeler:
PR #612 by @0001vrn supersedes this PR.

PR #610, opened 2019-11-04
Add trusted flag to detector

swfortier:
- Adds new trusted flag as an indicator on the output anomalies wether the detector is one that has been verified and trusted as valid. 
- New API allows enabling/disabling trusted flag by detector uuid.
- Updated Create Detector and Toggle Detector APIs to modify metadata and properly set last updated time.

### DIFF
--- a/anomdetect/src/main/java/com/expedia/adaptivealerting/anomdetect/detect/breakout/algo/edmx/EdmxDetector.java
+++ b/anomdetect/src/main/java/com/expedia/adaptivealerting/anomdetect/detect/breakout/algo/edmx/EdmxDetector.java
@@ -37,11 +37,14 @@ public final class EdmxDetector implements BreakoutDetector {
     private UUID uuid;
 
     @Getter
+    private boolean trusted;
+
+    @Getter
     private EdmxHyperparams hyperparams;
 
     private final EvictingQueue<MetricData> buffer;
 
-    public EdmxDetector(UUID uuid, EdmxHyperparams hyperparams) {
+    public EdmxDetector(UUID uuid, EdmxHyperparams hyperparams, boolean trusted) {
         notNull(uuid, "uuid can't be null");
         notNull(hyperparams, "hyperparams can't be null");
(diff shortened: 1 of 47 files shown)


anomdetect/src/main/java/com/expedia/adaptivealerting/anomdetect/detect/breakout/algo/edmx/EdmxDetector.java
```diff
@@ -59,6 +63,7 @@ public DetectorResult detect(MetricData metricData) {
 
         val warmup = buffer.remainingCapacity() > 0;
         val result = new EdmxDetectorResult().setWarmup(warmup);
+        val trusted = isTrusted();
```
williewheeler: Is there a point to creating a local variable here?

docker-compose/README.md
```diff
@@ -1,21 +1,26 @@
 # Testing the Adaptive Alerting pipeline with Docker Compose
 
-The Docker Compose setup allows you run the Adaptive Alerting components and test sample metrics against the project. By default the images are downloaded from Docker Hub, but they can be substituted with your own builds or different versions from Docker Hub for testing purposes.
+The Docker Compose setup allows you run the Adaptive Alerting components and test sample metrics against the project. You can begin by building the project's docker images through the Makefile. You can also subsitute previous versions by updating the docker-compose.yml file and referencing images from Docker Hub.
```
williewheeler: s/subsitute/substitute

anomdetect/src/main/java/com/expedia/adaptivealerting/anomdetect/detect/breakout/algo/edmx/EdmxDetector.java
swfortier: Not entirely necessary. It is used below in the return result. It can be `.setTrusted(isTrusted())` instead, but I was following the existing pattern.

PR #611, opened 2019-11-05
Added a check for empty list

vvishwakarma-hc:


modelservice/src/main/java/com/expedia/adaptivealerting/modelservice/repo/impl/DetectorRepositoryImpl.java
```diff
@@ -141,7 +141,11 @@ public DetectorDocument findByUuid(String uuid) {
         val queryBuilder = QueryBuilders.termQuery("uuid", uuid);
         val searchSourceBuilder = elasticsearchUtil.getSourceBuilder(queryBuilder).size(DEFAULT_ES_RESULTS_SIZE);
         val searchRequest = elasticsearchUtil.getSearchRequest(searchSourceBuilder, DETECTOR_INDEX, DETECTOR_DOC_TYPE);
-        return getDetectorsFromElasticSearch(searchRequest).get(0);
+        List<DetectorDocument> detectorsFromElasticSearch = getDetectorsFromElasticSearch(searchRequest);
+        if (detectorsFromElasticSearch.isEmpty()) {
+            throw new RuntimeException("No detector document found for uuid "+ uuid);
+        }
```
williewheeler: We need a way to distinguish the "not found" case from a more generic "something broke" case. Consider throwing something along the lines of `NotFoundException` or similar. That way we can return an HTTP 404 at the web tier.

PR #614, opened 2019-11-15
Issue 558: Adding a windowed simple moving average forecaster.

bcorbett:
The `SmaPointForecaster` is a new forecaster that calculates a moving average over a window whose length is specified in the forecaster's parameters. No warm-up is required. 
The forecaster can be started from scratch or from an initial set of data.
Due to the nature of incremental fixed-window SMA calculation, if an initial set of data points are specified, then they must include one FULL period's worth of data.


anomdetect/src/main/java/com/expedia/adaptivealerting/anomdetect/forecast/point/algo/sma/SmaPointForecaster.java
```diff
@@ -0,0 +1,84 @@
...
+    @Generated // https://reflectoring.io/100-percent-test-coverage/
+    private SmaPointForecasterParams params;
+
+    @Getter
+    @Generated // https://reflectoring.io/100-percent-test-coverage/
+    private double mean;
+
+    @Getter
+    @Generated // https://reflectoring.io/100-percent-test-coverage/
+    private EvictingQueue<Double> periodOfValues;
```
bcorbett: With an `EvictingQueue`, we are guaranteed that the length of the queue will never exceed the size set at construction time. This is a perfect data structure for maintaining a sliding window of values. 
We must maintain the values in our SMA's period due to the nature of incrementally calculating an SMA. As we'll see later, we need to remove the head's contribution to the mean and then add-in the newly observed value's contribution.

anomdetect/src/test/java/com/expedia/adaptivealerting/anomdetect/forecast/point/algo/sma/SmaPointForcasterTest.java
```diff
@@ -0,0 +1,117 @@
...
+        // [2.0, 4.0]; mean=3.0
+        assertEquals(3.0, sma3.forecast(createMetricData(4.0)).getValue(), TOLERANCE);
+        // [2.0, 4.0, 6.0]; mean=4.0
+        assertEquals(4.0, sma3.forecast(createMetricData(6.0)).getValue(), TOLERANCE);
+        // [4.0, 6.0, 8.0]; mean=6.0
+        assertEquals(6.0, sma3.forecast(createMetricData(8.0)).getValue(), TOLERANCE);
+    }
+
+    @Test
+    public void testAgainstSampleData() {
```
bcorbett: Our sample data in the CSV has pre-calculated values for SMA(1), SMA(3), SMA(9), and SMA(21). This one method will test all of these forecasters at the same time in a single iteration over the dataset.

williewheeler:
Thanks for the PR @bcorbett. I expect to review this sometime next week.

bcorbett:
@williewheeler by the way, if you think that having the caller supply an initial full period of values when the forecaster is being seeded is too difficult or problematic, then there is another approach we could use. 
We could adopt the same pattern the other forecasters use, which is to specify a single `initialMean` value. 
If specified, the forecaster would create its internal queue (of length `lookBackPeriod `) and just fill in `initialMean` for all of the values. So if `initialMean` was 33.5 and `lookBackPeriod was 3, the queue would be initialized to `[33.5, 33.5, 33.5]`. It's not ideal because it's unlikely that those were the actual values, but as the number of inputs grows the impact of the initial error obviously shrinks.

williewheeler:
@bcorbett Note that I haven't forgotten about this PR. Will review shortly.

williewheeler:
@bcorbett I've merged your changes. Thank you very much for the contribution.

Regarding initialization, this is an issue that confronts other detectors as well. We are going to have to do something where on startup, a detector automatically self-initializes by pulling data from some metric data store. The reason is that we don't want to lose our runtime detector initialization every time we redeploy the Detector Manager. For SMA with short periods it may not be that big of a deal, but for something like a seasonal naive forecaster with for a metric with a weekly seasonality, we'd have to wait a whole week before the detector kicks in, which isn't any good. So I'd expect us to do the same thing with this detector.

PR #623, opened 2019-11-28
Log exceptions in splunk for ad-mapper

karanshah03:


karanshah03:
Not required anymore.

PR #626, opened 2019-11-29
adding counter for exception in custom exception handler.

tusharbahl:


modelservice/src/main/java/com/expedia/adaptivealerting/modelservice/web/DetectorController.java
```diff
@@ -65,6 +69,7 @@ public DetectorDocument findByUuid(@RequestParam String uuid) {
     public List<DetectorDocument> findByCreatedBy(@RequestParam String user) {
         List<DetectorDocument> detectors = detectorRepo.findByCreatedBy(user);
         if (detectors == null || detectors.isEmpty()) {
+            exceptionCounter.increment(1.0);
```
karanshah03: @tusharbahl This doesn't track the exceptions which are thrown in detector repo and detector mapping impl classes.

williewheeler:
Two things:

1. Please resolve the merge conflicts.
2. Consider using a ControllerAdvice here. See https://medium.com/@jovannypcg/understanding-springs-controlleradvice-cd96a364033f. That way we count all exceptions and we don't have to keep adding counters throughout the controller code.

modelservice/src/main/java/com/expedia/adaptivealerting/modelservice/exception/CustomExceptionHandler.java
```diff
@@ -17,22 +19,27 @@
 @Generated
 public class CustomExceptionHandler extends ResponseEntityExceptionHandler {
 
+    private Counter exceptionCounter = Metrics.counter("exceptionCounter");
+
     @ExceptionHandler(Exception.class)
     public final ResponseEntity<Object> handleAllExceptions(Exception ex, WebRequest request) {
```
williewheeler: Don't need `request` here since you aren't using it. Same below.

PR #628, opened 2019-12-11
move aa-metric-functions input file deployment to load from local pat…

tusharbahl:
…h instead of configmap


tusharbahl:
moving aa-metric-functions input file deployment to load from local path instead of configmap.

PR #630, opened 2019-12-12
add config to use initContainers to download input file for aa-metric…

dinilatgit:
…-functions


deployment/terraform/aa-metric-functions/templates/deployment_yaml.tpl
```diff
@@ -62,12 +62,21 @@ spec:
 #          initialDelaySeconds: 30
 #          periodSeconds: 5
 #          failureThreshold: 6
+      # Add initContainer to download input file to /config/functions.txt
+      initContainers:
+      - name: download-input-file
+        # example - image: busybox
+        image: "${initContainer_image}"
+        # example - command: ["sh", "-c", "curl", "<input-file-location>", "-o", "/config/functions.txt"]
```
hridyeshpant: can we remove this line?

PR #634, opened 2019-12-16
adding logging for the exception in modelservice.

tusharbahl:


modelservice/src/main/java/com/expedia/adaptivealerting/modelservice/exception/CustomExceptionHandler.java
```diff
@@ -47,19 +49,18 @@ public CustomExceptionHandler(MeterRegistry meterRegistry) {
         List<String> details = new ArrayList<>();
         details.add(ex.getLocalizedMessage());
         ErrorResponse error = new ErrorResponse("Server Error", details);
+        log.error("Exception occurred "+ error);
```
mattcallanan: Please change this to `log.error("Unhandled exception encountered, responding with internal server error", error);`

See http://www.slf4j.org/apidocs/org/slf4j/Logger.html#error(java.lang.String,java.lang.Throwable)

PR #642, opened 2019-12-23
Add SeasonalNaiveDetector

mattcallanan:


anomdetect/src/test/java/com/expedia/adaptivealerting/anomdetect/detect/outlier/algo/forecasting/LegacySeasonalNaiveFactoryTest.java
```diff
@@ -0,0 +1,53 @@
...
+    private static final int ONE_MINUTE_IN_SECONDS = 60;
+
+    @Test
+    public void testBuildDetector() {
+        val factoryUnderTest = new LegacySeasonalNaiveDetectorFactoryProvider();
+        val document = readDocument("seasonal-naive");
+        val detector = factoryUnderTest.buildDetector(document);
+        val seasonalNaive = (SeasonalNaivePointForecaster) detector.getPointForecaster();
+        val seasonalNaiveParams = seasonalNaive.getParams();
+        val welford = (ExponentialWelfordIntervalForecaster) detector.getIntervalForecaster();
```
williewheeler: Welford won't work for seasonal series. You suggested a modified version of the Welford for Holt-Winters, but that modification won't work for the seasonal naive case because there's only a single cycle (so variance is always 0).

My suggestion for the short term is to use a multiplicative interval forecaster. Longer term we will need some other kind of interval forecaster here -- maybe we can modify the power law interval forecaster to work in an online fashion (that should be easy enough to do). Another possibility is to add a new interval forecaster that estimates variance based on a local window around the current timestamp, and uses that to generate a threshold.

PR #643, opened 2019-12-23
Exception handling/modelservice

tusharbahl:


tusharbahl:
RecordNotFoundException is also throwing 404, earlier it was protected access modifier due to which it wasn't getting handled properly. Closing this PR and opening another one in just making it public instead of protected in CustomExceptionHandler

PR #661, opened 2020-01-28
Add warmup to seasonal naive

tkamenov-expedia:
Add warmup to seasonal naive

### DIFF
--- a/anomdetect/src/test/java/com/expedia/adaptivealerting/anomdetect/detect/outlier/algo/forecasting/ForecastingDetectorTest.java
+++ b/anomdetect/src/test/java/com/expedia/adaptivealerting/anomdetect/detect/outlier/algo/forecasting/ForecastingDetectorTest.java
@@ -37,6 +37,7 @@ import java.util.UUID;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNotNull;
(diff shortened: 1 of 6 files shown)


anomdetect/src/test/java/com/expedia/adaptivealerting/anomdetect/detect/outlier/algo/forecasting/ForecastingDetectorTest.java
```diff
@@ -94,6 +95,7 @@ public void testDetect_nullForecast() {
 
         val result = (OutlierDetectorResult) detectorUnderTest.detect(metricData);
         assertNotNull(result);
+        assertFalse(result.isWarmup());
```
mattcallanan: Can we also add a test case for `assertTrue(result.isWarmup())`?

Something like:

```
    @Test
    public void testDetect_warmingUpForecast() {
        initDependencies(true);
        val metricDef = TestObjectMother.metricDefinition();
        val metricData = new MetricData(metricDef, 100.0, Instant.now().getEpochSecond());
        val result = (OutlierDetectorResult) detectorUnderTest.detect(metricData);
        assertTrue(result.isWarmup());
        assertNotNull(result);
    }
```

Above code assumes parameterising the `warmup` setting in `initDependencies()` and calling `initDependencies()` from every test case instead of in `setUp()`.

PR #665, opened 2020-02-06
Throttling for Data Initialization

mattcallanan:


anomdetect/src/main/java/com/expedia/adaptivealerting/anomdetect/source/data/initializer/throttlegate/RandomThrottleGate.java
```diff
@@ -0,0 +1,44 @@
...
+ * limitations under the License.
+ */
+package com.expedia.adaptivealerting.anomdetect.source.data.initializer.throttlegate;
+
+import lombok.extern.slf4j.Slf4j;
+
+import java.util.Random;
+
+@Slf4j
+public class RandomThrottleGate implements ThrottleGate {
```
mattcallanan: Please write a unit test for this.
Something that calls isOpen() 1,000 times and asserts that it returns `true` around `1000 * throttleGateLikelihood` of the time (within an acceptable range) would be good.

PR #670, opened 2020-02-12
Renamed graphite to metric tank and updated http client wrapper to su…

karanshah03:
Renamed graphite source to metric tank and updated HTTP client wrapper to support customizable headers.

Later on, we would like to support both graphite & metric tank data sources by moving code to some core package and then reuse it for both graphite & metric tank adapter

### DIFF
--- a/anomdetect/src/main/java/com/expedia/adaptivealerting/anomdetect/source/data/metrictank/MetricTankClient.java
+++ b/anomdetect/src/main/java/com/expedia/adaptivealerting/anomdetect/source/data/metrictank/MetricTankClient.java
@@ -22,7 +22,6 @@ import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
 import lombok.val;
 import org.apache.http.client.fluent.Content;
-import org.apache.http.client.fluent.Request;
 
 import java.io.IOException;
 import java.util.Arrays;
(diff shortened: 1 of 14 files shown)


anomdetect/src/main/java/com/expedia/adaptivealerting/anomdetect/source/data/metrictank/MetricTankClient.java
```diff
@@ -53,35 +55,43 @@
...
         }
 
-        List<GraphiteResult> results;
+        List<MetricTankResult> results;
         try {
-            results = Arrays.asList(objectMapper.readValue(content.asBytes(), GraphiteResult[].class));
+            results = Arrays.asList(objectMapper.readValue(content.asBytes(), MetricTankResult[].class));
         } catch (IOException e) {
-            throw new GraphiteClientException(String.format("IOException while parsing response from Graphite target: %s", target), e);
+            throw new MetricTankClientException(String.format("IOException while parsing response from Graphite target: %s", target), e);
```
williewheeler: s/Graphite/Metrictank

anomdetect/src/main/java/com/expedia/adaptivealerting/anomdetect/source/data/metrictank/MetricTankClient.java
```diff
@@ -53,35 +55,43 @@
...
         }
         return results;
     }
+
+    private Map<String, String> buildHeaders() {
+        val headers = new HashMap<String, String>();
+        headers.put("x-org-id", "1");
+        return headers;
```
williewheeler: Does this map need to be mutable? If not, prefer singletonMap:
https://stackoverflow.com/questions/7125536/when-would-i-use-java-collections-singletonmap-method

karanshah03: good call. updating it to singletonMap

PR #677, opened 2020-02-14
Enhancement/snap to interval length

karanshah03:


karanshah03:
Too many conflicts. Will raise a new one.

PR #692, opened 2020-02-28
Bug fix/wrong body format

karanshah03:


karanshah03:
Raising a new one. There is something wrong with this one. Travis is complaining error for line which doesn't even exist.

PR #712, opened 2020-04-13
Visualizer

pranavsundriyal:


kafka/src/main/java/com/expedia/adaptivealerting/kafka/visualizer/AnomalyModel.java
```diff
@@ -0,0 +1,75 @@
+package com.expedia.adaptivealerting.kafka.visualizer;
+
+import com.expedia.adaptivealerting.anomdetect.detect.AnomalyThresholds;
+import com.expedia.metrics.TagCollection;
+
+public class AnomalyModel {
```
williewheeler, line 6: Please use Lombok's `@Data` annotation here (as opposed to explicitly defining all the accessors), and please make the member fields private.

kafka/src/main/java/com/expedia/adaptivealerting/kafka/visualizer/Utility.java
```diff
@@ -0,0 +1,30 @@
+package com.expedia.adaptivealerting.kafka.visualizer;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+import java.text.SimpleDateFormat;
+import java.util.Date;
+
+public class Utility {
```
williewheeler, line 9: Please choose a classname that clarifies what sort of utility this is. Also please add Lombok `@UtilityClass`

kafka/src/main/java/com/expedia/adaptivealerting/kafka/visualizer/Utility.java
```diff
@@ -0,0 +1,30 @@
...
+
+    public static String convertToJson(Object object) {
+        String json = "";
+        ObjectMapper objectMapper = new ObjectMapper();
+        try {
+            json = objectMapper.writeValueAsString(object);
+        } catch (JsonProcessingException e) {
+            e.printStackTrace();
+        }
+        return json;
```
williewheeler, line 20: If the exception occurs, then this will return an empty string, which isn't valid JSON. Better to simply throw an exception if the conversion fails since the client will have more context to respond. Feel free to wrap with a `RuntimeException` if you don't want to throw the original.

kafka/src/test/java/com/expedia/adaptivealerting/kafka/AnomalyVisualizerTest.java
```diff
@@ -0,0 +1,32 @@
...
+import com.expedia.adaptivealerting.kafka.visualizer.AnomalyModel;
+import org.elasticsearch.client.Response;
+import org.junit.Before;
+import org.junit.Test;
+
+import java.util.Date;
+
+import static org.junit.Assert.assertNotNull;
+
+public class AnomalyVisualizerTest {
```
williewheeler, line 12: Please implement the tests.

kafka/src/main/java/com/expedia/adaptivealerting/kafka/visualizer/ElasticSearchService.java
```diff
@@ -0,0 +1,48 @@
...
+public class ElasticSearchService implements Runnable {
+
+    private static String HOST = "localhost";
+    private static int PORT = 9200;
+    private static String SCHEME = "http";
+    private static String METHOD = "POST";
+    private static String END_POINT = "/anomalies/doc";
+    private Object object;
+
+    public ElasticSearchService(Object Object) {
```
williewheeler, line 23: What is this object? Either Javadoc or choose a clearer name. Also the param name should use standard Java conventions -- lowercase name.

kafka/src/main/java/com/expedia/adaptivealerting/kafka/visualizer/ElasticSearchService.java
```diff
@@ -0,0 +1,48 @@
...
+        this.object = object;
```
williewheeler, line 24: This looks like a bug -- the param name is `Object`, so it looks like you are assigning the member to itself.

PR #714, opened 2020-04-21
Enhancement/external detectors

karanshah03:
This change makes integration with external detectors seamless. `KafkaAnomalyDetectorMapper` will now push messages onto `type-specific` topics where they can be picked up by the internal/external detectors. 

- For internal detectors, `ad-manager` is the consumer and all the mapped metrics are routed to `defaultOutputTopic` for its consumption.

- For external detectors, all the messages will now be routed to their own Kafka topics i.e. `defaultOutputTopic + "-" + consumerId`

Let's say we want to add a new external detector `rcf`. We now just need to create a new detector mapping like below in AA DB and all the metrics mapped to rcf now will be routed to its own Kafka topic i.e. `defaultOutputTopic + "-" + "rcf"`. It can then generate anomalies from these mapped metrics and map it to the output format that AA expects, and push the output to the AA anomalies Kafka topic.

``` {
	"detector": {
		"uuid": "e28c351e-5863-4e15-8bb0-f4cd0e6a2843",
		"consumerId": "rcf"
	},
	"expression": {
		"operator": "AND",
		"operands": [{
				"field": {
					"key": "app-name",
					"value": "test-app"
				},
				"expression": null
			}
		]
	},
	"user": {
		"id": "userid"
	}
}

### CODE REVIEW

File: anomdetect/src/test/resources/groupedDetectors.json [hunk 1]
```diff
@@ -2,21 +2,25 @@
...
+        "consumerId": "cid",
         "uuid": "2c49ba26-1a7d-43f4-b70c-c6644a2c1689"
       }
     ],
     "4": [
       {
+        "consumerId": "cid",
         "uuid": "5eaa54e9-7406-4a1d-bd9b-e055eca1a423"
       }
     ]
```
hridyeshpant, line 26: please add test with no consumerId and see if that is passing . I doubt about when we are doing 
things like result.add(detector.getConsumerId() + "," + detector.getUuid().toString()); with null values

anomdetect/src/main/java/com/expedia/adaptivealerting/anomdetect/mapper/CacheUtil.java
```diff
@@ -33,47 +33,73 @@
 @Slf4j
 public final class CacheUtil {
 
-    private static final String DELIMITER = "->";
+    private static final String CACHE_KEY_DELIMITER = "->";
+    private static final String CACHE_VALUE_DELIMITER = "|";
 
+    /**
```
williewheeler, line 39: Thanks for adding all the Javadoc. Very helpful.

anomdetect/src/main/java/com/expedia/adaptivealerting/anomdetect/mapper/DetectorMapperCache.java
```diff
@@ -104,11 +103,11 @@ private MetricSet metricsFor(String cacheName, final Cache cache) {
      * @return the list of Detectors
      */
     public List<Detector> get(String key) {
-        String bunchOfCachedDetectorIds = cache.getIfPresent(key);
-        if (bunchOfCachedDetectorIds == null) {
+        String value = cache.getIfPresent(key);
+        if (value == null) {
             return Collections.emptyList();
```
williewheeler, line 108: The old variable name is clearer to me. :D

anomdetect/src/main/java/com/expedia/adaptivealerting/anomdetect/mapper/DetectorMapperCache.java
```diff
@@ -117,59 +116,59 @@ private MetricSet metricsFor(String cacheName, final Cache cache) {
...
-        List<Detector> detectors = CacheUtil.buildDetectors(detectorIdsString);
-        detectorUuids.forEach(uuid -> {
-            detectors.remove(new Detector(uuid));
+    private String removeDisabledDetectors(List<Detector> detectorsToBeDisabled, String value) {
+        List<Detector> detectorsStoredInCache = CacheUtil.buildDetectors(value);
+        detectorsToBeDisabled.forEach(detectorToBeDisabled -> {
+            detectorsStoredInCache.remove(new Detector(detectorToBeDisabled.getConsumerId(), detectorToBeDisabled.getUuid()));
         });
-        return CacheUtil.getDetectorIds(detectors);
+        return CacheUtil.getValue(detectorsStoredInCache);
```
williewheeler, line 171: Similar comments as I made above - `getValue()` is much less clear than `getDetectorIds()`. Please reconsider the names here.

PR #717, opened 2020-04-23
Detector mapper bloom filter

brett-miller:


brett-miller:
This still needs a bit of work.  Leaving the PR open for visibility.

PR #722, opened 2020-04-29
adding deployment configs for visualizer

pranavsundriyal:


deployment/terraform/visualizer/main.tf
```diff
@@ -0,0 +1,76 @@
+locals {
+  app_name = "ad-mapper"
```
williewheeler, line 2: shouldn't this be visualizer?

deployment/terraform/visualizer/main.tf
```diff
@@ -0,0 +1,76 @@
...
+  config_file_path = "${path.module}/templates/ad-mapper_conf.tpl"
```
williewheeler, line 3: here too -- there are some others

deployment/terraform/visualizer/main.tf
```diff
@@ -0,0 +1,76 @@
...
+  count = "${var.enabled?1:0}"
+  checksum = "${sha1("${data.template_file.config_data.rendered}")}"
+  configmap_name = "ad-mapper-${local.checksum}"
+}
+
+data "template_file" "config_data" {
+  template = "${file("${local.config_file_path}")}"
+  vars {
+    kafka_endpoint = "${var.kafka_endpoint}"
+    modelservice_base_uri = "${var.modelservice_base_uri}"
```
williewheeler, line 14: does the visualizer reference the model service?

deployment/terraform/main.tf
```diff
@@ -162,3 +162,36 @@ module "aa-metric-functions" {
...
+  kubectl_executable_name = "${var.kubectl_executable_name}"
+  kubectl_context_name = "${var.kubectl_context_name}"
+
+  # Environment
+  jvm_memory_limit = "${var.visualizer["jvm_memory_limit"]}"
+  graphite_hostname = "${var.graphite_hostname}"
+  graphite_port = "${var.graphite_port}"
+  graphite_enabled = "${var.graphite_enabled}"
+  graphite_prefix = "${var.graphite_prefix}"
+  env_vars = "${var.ad-mapper["environment_overrides"]}"
```
williewheeler, line 191: ad-mapper?

deployment/terraform/visualizer/templates/main.tf
```diff
@@ -0,0 +1,81 @@
+locals {
+  app_name = "aa-metric-functions"
```
hridyeshpant, line 2: shouldn't this be visualizer?

deployment/terraform/visualizer/templates/main.tf
```diff
@@ -0,0 +1,81 @@
...
+  config_file_path = "${path.module}/templates/aa-metric-functions_conf.tpl"
```
hridyeshpant, line 3: please change aa-metric-functions_conf.tpl to visualizer_conf.tpl

deployment/terraform/visualizer/main.tf
```diff
@@ -0,0 +1,76 @@
...
+    jvm_memory_limit = "${var.jvm_memory_limit}"
+    graphite_enabled = "${var.graphite_enabled}"
+    graphite_host = "${var.graphite_hostname}"
+    graphite_port = "${var.graphite_port}"
+    graphite_prefix = "${var.graphite_prefix}"
+    env_vars = "${indent(9,"${var.env_vars}")}"
+  }
+}
+
+resource "kubernetes_config_map" "ad-mapper-config" {
```
hridyeshpant, line 48: ad-mapper?

deployment/terraform/visualizer/templates/main.tf
```diff
@@ -0,0 +1,81 @@
...
+  deployment_yaml_file_path = "${path.module}/templates/deployment_yaml.tpl"
+  count = "${var.enabled?1:0}"
+  checksum = "${sha1("${data.template_file.config_data.rendered}")}"
+  configmap1_name = "aa-metric-functions-config-${local.checksum}"
```
hridyeshpant, line 7: Do we need configmap1_name for visualizer ?

deployment/terraform/visualizer/templates/main.tf
```diff
@@ -0,0 +1,81 @@
...
+    graphite_host = "${var.graphite_hostname}"
+    graphite_port = "${var.graphite_port}"
+    graphite_prefix = "${var.graphite_prefix}"
+    initContainer_image = "${var.initContainer_image}"
+    download_input_file_command = "${var.download_input_file_command}"
+    env_vars = "${indent(9,"${var.env_vars}")}"
+  }
+}
+
+ resource "kubernetes_config_map" "aa-metric-functions-config" {
```
hridyeshpant, line 53: shouldn't this be visualizer?

deployment/terraform/visualizer/templates/main.tf
```diff
@@ -0,0 +1,81 @@
...
+  metadata {
+    name = "${local.configmap1_name}"
+    namespace = "${var.namespace}"
+  }
+  data {
+    "aa-metric-functions.conf" = "${data.template_file.config_data.rendered}"
```
hridyeshpant, line 59: shouldn't this be visualizer?

deployment/terraform/visualizer/templates/visualizer_conf.tpl
```diff
@@ -0,0 +1,12 @@
+visualizer {
+    metric-consumer {
+      bootstrap.servers = "localhost:19092"
```
williewheeler, line 3: Bootstrap servers should use `${kafka_endpoint}` here, not a hardcoded host/port.

deployment/terraform/visualizer/templates/visualizer_conf.tpl
```diff
@@ -0,0 +1,12 @@
...
+      group.id = "my-group"
+      topic = "anomalies"
+      key.deserializer = "org.apache.kafka.common.serialization.StringDeserializer"
+      value.deserializer = "com.expedia.adaptivealerting.kafka.serde.MappedMetricDataJsonSerde$Deser"
+    }
+    elastic-search {
+      host = "localhost"
```
williewheeler, line 10: Similarly I would expect this to be controlled by a variable, not by a hardcode.

williewheeler: Really all of these values should be controlled by variables. The example from `ad-manager_conf.tpl` is much better than `ad-mapper_conf.tpl`.

deployment/terraform/visualizer/templates/visualizer_conf.tpl
```diff
@@ -0,0 +1,12 @@
+visualizer {
+    metric-consumer {
+      bootstrap.servers = "${kafka_endpoint}"
+      group.id = "my-group"
+      topic = "anomalies"
+      key.deserializer = "org.apache.kafka.common.serialization.StringDeserializer"
+      value.deserializer = "com.expedia.adaptivealerting.kafka.serde.MappedMetricDataJsonSerde$Deser"
```
williewheeler, line 7: These should all be configurable

pranavsundriyal: i was checking metric function app and seems like not all configs need to come from deployer since these configs dont change based on environment is prod or test. 
https://github.com/ExpediaDotCom/adaptive-alerting/blob/master/deployment/terraform/aa-metric-functions/templates/aa-metric-functions_conf.tpl#L5

williewheeler: In principle those ought to be configurable per env actually. For Expedia we probably won't change the serializers anytime soon, but since you will be updating the group ID and the topic name (the topic name for sure needs to be configurable), it makes sense to just make the serializers configurable too. I know metric-functions isn't doing it, but I would ask that we make visualizer do it the right way so we can use visualizer as the gold standard instead of just copying the hardcodes in metric-functions.

williewheeler:
I've approved the changes - thanks @pranavsundriyal. I suggest waiting til we have the deployer ready before merging these.

PR #741, opened 2020-05-19
Bug fix for elastic search url

karanshah03:
Bug fix for elastic search url. We don't use host and port but the entire URL as HttpHost


modelservice/src/main/java/com/expedia/adaptivealerting/modelservice/elasticsearch/ElasticSearchProperties.java
```diff
@@ -47,27 +46,4 @@
     }
 
     private Config config;
```
hridyeshpant, line 48: i think the name of this variable makes more sense to me as `EsProperties`, not config but if that makes too many place changes then leave it. we already have ElasticsearchConfig.java and then we ar using `ElasticSearchProperties.Config config;` , we are using config and Properties. i think the config where we are setting the configuration value for  ES Properties

PR #742, opened 2020-05-20
Enhancement/detector last used

karanshah03:
Upgraded detector path from v2 to v3 and updated detector manager to update detector's `lastAccessedTime` every time it is invoked to keep track of detector usage.

### DIFF
--- a/anomdetect/src/main/java/com/expedia/adaptivealerting/anomdetect/DetectorManager.java
+++ b/anomdetect/src/main/java/com/expedia/adaptivealerting/anomdetect/DetectorManager.java
@@ -276,14 +276,14 @@ public class DetectorManager {
      */
     void detectorLastUsedTimeSync(long currentTime) {
         long updateDurationInSeconds = (currentTime - detectorsLastUsedSyncedTillTime) / 1000;
-        int queueSize = detectorsLastUsedTimeToBeUpdated.size();
+        int detectorToBeUpdatedQueueSize = detectorsLastUsedTimeToBeUpdated.size();
 
-        if (updateDurationInSeconds <= 0 || queueSize <= 3) {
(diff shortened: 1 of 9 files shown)


anomdetect/src/main/java/com/expedia/adaptivealerting/anomdetect/DetectorManager.java
```diff
@@ -261,7 +266,28 @@ private void checkMappedMetricData(MappedMetricData mappedMetricData) {
...
+        if (updateDurationInSeconds <= 0 || detectorsLastUsedTimeToBeUpdated.isEmpty()) {
+            return;
+        }
+
+        int queueSize = detectorsLastUsedTimeToBeUpdated.size();
+        //Update last used time only for 1/4th of queue's size at one go for better performance
+        for (int i = 0; i < queueSize / 4; i++) {
+            UUID uuid = detectorsLastUsedTimeToBeUpdated.remove();
+            if (uuid != null) {
+                detectorSource.updatedDetectorLastUsed(uuid);
```
hridyeshpant, line 288: but still, we are making a single call per UUID? 
Can we use bulk API call and construct one single request for 1/4 detectors at a time?

karanshah03: yeah. I already added a comment to explore on that option. we can do this in second pass.

karanshah03: also, update calls are made via scheduler now and not for every metric that comes in.

karanshah03: also, update calls are made via scheduler now and not for every metric that comes in.

PR #743, opened 2020-05-20
Initial commit for detector runner

pranavsundriyal:
@pranavsundriyal can you please add some more description about this PR.
Right now I am not able to understand the use case for all these changes.
What is the problem statement and why we are doing these changes and how it will help?

@karanshah03 can you review this as well


pranavsundriyal:
Creating this springboot repo as a way for other team to add their on algos/detector and run them on existing kafka data and write back anomalies to kafka queue. Eventually this app will not be in the adaptive-alerting repo and rather be forked out in the expedia internal by other teams for use or it can be probably broken down into a lib which can be used by any team that has a algo they want to test. @williewheeler would you like to add more to the description.

PR #762, opened 2020-07-02
remove tracing for KafkaAnomalyDetectorMapper as source library repor…

dinilatgit:
…ts internal error with use


dinilatgit:
`Exception in thread "main" java.lang.IllegalAccessError: tried to access class io.opentracing.contrib.kafka.SpanDecorator from class io.opentracing.contrib.kafka.streams.TracingKafkaClientSupplier`

error is reported when using the library for tracing KStreams.

PR #765, opened 2020-07-06
add finish to detectormanager span

dinilatgit:


dinilatgit:
Adding the detector manager span only to add tags to the complete trace which is created and managed by the library. The span is not designed to calculate trace in detector manager, that is done by the library introduced interceptor class itself.

https://github.com/opentracing-contrib/java-kafka-client/issues/77 has been logged to know if there is a way to provide tags to the library functions. Detector Manager span is a work around.

